PR #8, opened 2024-11-05
Start of updating styles

rcubbon:


src/app/ClientLayout.tsx
```diff
@@ -0,0 +1,20 @@
+// ClientLayout.tsx
```
danbastin, line 1: I like pulling this out into it's own layout 👍

PR #17, opened 2024-11-23
Google Auth is set up

wanderwayward:
You need to have the secret and clientid in your .env.local

do I paste them here or google chat?


wanderwayward:
Discord log in is now set up too.

PR #21, opened 2024-11-28
fix: hide opposing player's cards

farisT:


danbastin:
I'll need to see if we can add this to the linter, but I'd like to stay consistent with the engine repo in keeping our indentation to 4 spaces. Right now they are tabs so I'll see if I can get the formatter to switch that over. 

As for the card visibility: This is actually something that gets controlled on the server side in the card summary function. We'll want to utilize this so that we remain tightly integrated with the information we're getting from the server and make sure that is being handled appropriately on the front end. Let me know in discord if you have any questions about this.

PR #25, opened 2024-12-10
Che bato/custom decks

CheBato:
- Renamed S3Assets to S3Utils
- Added utility functions to convert deck data from SWUDB to deck data compatible with the server. Located in the s3Utils.ts
- Added a new GET method in the API/S3Bucket/route.ts to receive json files from the S3Bucket.


src/app/_utils/s3Utils.ts
```diff
@@ -0,0 +1,194 @@
+import { CardData } from "../_components/_sharedcomponents/Cards/CardTypes";
+
+// Deck data interfaces for deck info from swudb
+interface DeckMetadata {
```
AMMayberry1, line 4: nit: the convention we use in the forceteki repo is that all interface and type names start with "I" (e.g. IDeckMetadata), not sure if y'all want to use the same convention here

CheBato: that might be a good convention to have

src/app/_utils/s3Utils.ts
```diff
@@ -0,0 +1,194 @@
...
+        } else {
+            // If no mapping found, log a warning and retain the original ID
+            console.warn(`No internalName mapping found for ID: ${id}`);
+            return { id, count };
+        }
+    };
+
+    // Update leader, secondleader, base
+    const updatedLeader = updateCardSlot(deckData.leader);
+    let updatedSecondLeader = null;
```
danbastin, line 126: 😉

src/app/_components/_sharedcomponents/CreateGameForm/CreateGameForm.tsx
```diff
@@ -60,7 +61,7 @@ const CreateGameForm: React.FC<CreateGameFormProps> = ({
 		useState<string>("Vader Green Ramp");
 	const [deckLink, setDeckLink] = useState<string>("");
 	const [saveDeck, setSaveDeck] = useState<boolean>(false);
-	const [deckData, setDeckData] = useState<DeckData | null>(null);
+	//let [deckData, setDeckData] = useState<DeckData | null>(null); Because of Async this won't set in the correct timeframe
```
danbastin, line 64: This was just a temporary holding spot for the data so we can get rid of this for now since the form has no reason to hold it.

src/app/_utils/s3Utils.ts
CheBato: changed the names of most of the interfaces. Except for those that do overrides.

danbastin (approved):

PR #26, opened 2024-12-15
Che bato/lobby design

CheBato:
- Divided the LeaderBaseCard to LeaderCard and BaseCard for more flexability
- Changed the design of the lobby to be similar to designs
- Added counter to the gameCards so we can see the amount of a certain card.
- Added Sideboard design and functionality to lobby.
- Changed the design of the leaders and base for the lobby.
- Changed how the Chat and Setup column looks in the lobby.
- Added API call to the server to update the deck information for the game (sideboarding).
- Added resizing of cards and elements in the lobby to support smaller screens.


src/app/_components/_sharedcomponents/CardArea/CardArea.tsx
```diff
@@ -3,23 +3,25 @@ import { Box } from "@mui/material";
 import GameCard from "../Cards/GameCard/GameCard";
 import { ICardAreaProps } from "@/app/_components/Gameboard/GameboardTypes";
 
-const CardArea: React.FC<ICardAreaProps> = ({ cards }) => {
+const CardArea: React.FC<ICardAreaProps> = ({ cards, pile }) => {
```
danbastin, line 6: I think we can get rid of this wrapper all together. It's not providing a lot of utility and by having components directly using the `GameCard` compnent they can better control the layout of the cards as well as which size and elements a cards should display.

src/app/_components/_sharedcomponents/Cards/LeaderCard/LeaderCard.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { useGame } from "@/app/_contexts/Game.context";
+import { s3CardImageURL } from "@/app/_utils/s3Utils";
+
+interface ILeaderCardProps {
+    isLobbyView?: boolean;
+    title?: string;
+    card: ICardData;
+}
+
+const LeaderCard: React.FC<ILeaderCardProps> = ({ isLobbyView = false, title, card }) => {
```
danbastin, line 13: I don't think these cards need to be split into two different components. They're quite similar to each other in most regards. Same as the GameCard what I'd rather see is that we keep the card as a LeaderBaseCard and have the props be the elements of the card that we want shown. In this case it would be The player nameplate and the damage counter. Otherwise these cards display the same way.

src/app/_components/_sharedcomponents/LeaderBaseBoard/LeaderBaseBoard.tsx
```diff
@@ -13,37 +13,61 @@ const LeaderBaseBoard: React.FC<ILeaderBaseBoardProps> = ({
...
-			{ isLobbyView ? null : <CardActionTray /> }
-			{/* Player's row */}
-			<Grid sx={rowStyle}>
-				<LeaderBase
-					player={connectedPlayer}
-					isLobbyView={isLobbyView}
-					title={connectedPlayer}
-				/>
-			</Grid>
+			{isLobbyView ? (
```
danbastin, line 29: This was the component that I was saying we should get rid of. I think having both players leader and base in a vertiacl colum like this is utlimately too restrictive. This may hold true for the LeaderBase component as well. By allowing each component direct access to the cards themselves we can be more flexible with them.

src/app/_components/_sharedcomponents/Cards/GameCard/GameCard.tsx
```diff
@@ -2,51 +2,112 @@ import React from "react";
...
+	const { sendGameMessage, connectedDeck, setConnectedDeck } = useGame();
 	const cardBorderColor = (card: ICardData) => {
 		if (card.selected) return "yellow";
 		if (card.selectable) return "limegreen";
 		if (card.exhausted) return "gray";
 		return "";
 	}
 
+	// handle deck movement
+	const handleSwitch = () => {
```
danbastin, line 40: This is Lobby specific logic that would be better suited living in the Lobby Deck component. In general this component should not care about what page or pile it belongs to. We wnat to tell the component how we would like it to render in those components and pass those options as props. If we want to change the onclick behavior of the component we could pass that as handler through the props. For the game card I think since in most cases this will be on the gameboard it makes sense to have the game message as it's default handler and then use the onclick handler provided in props as an override if one is provided.

src/app/_components/_sharedcomponents/Cards/GameCard/GameCard.tsx
```diff
@@ -31,22 +47,35 @@ const GameCard: React.FC<IGameCardProps> = ({
 	const styles = {
 		cardStyles: {
 			borderRadius: ".38em",
-			height: size === "standard" ? "10rem" : "8rem",
-			width: size === "standard" ? "7.18rem" : "8rem",
+			...(variant === "lobby"
+					? {
+						height: "18vh",
```
danbastin, line 52: In general we want to avoid using view sizes (vh, vw) for elements that aren't related to the page layout. We want the card size to be predicatble given the container that it's in and basing that size off the whole window makes that more difficult. Lets just make this a good default size using rem and then we can take a closer look at sizing when we do a responsive pass later.

src/app/_components/_sharedcomponents/Cards/LeaderBaseCard/LeaderBaseCard.tsx
```diff
@@ -13,116 +13,148 @@ import { s3CardImageURL } from "@/app/_utils/s3Utils";
...
-	card
-}) => {
-	const cardBorderColor = (card: ICardData) => {
-		if (!card) return "";
-		if (card.selected) return "yellow";
-		if (card.selectable) return "limegreen";
-		if (card.exhausted) return "gray";
-		return "black";
-	};
+                                                            variant,
```
danbastin, line 16: Clean up indentation here

src/app/_components/_sharedcomponents/Cards/LeaderBaseCard/LeaderBaseCard.tsx
```diff
@@ -13,116 +13,148 @@ import { s3CardImageURL } from "@/app/_utils/s3Utils";
...
-		fontWeight: "600",
-		fontSize: "1em",
-	};
+    const { sendGameMessage } = useGame();
 
-	const { sendGameMessage } = useGame();
+    return (
+        <Box>
+            {/* Show title above the card if in lobby view and variant is leader */}
+            {variant === "leader" && isLobbyView && title && (
```
danbastin, line 120: Similar to GameCard let's try to make thuis one more location agnostic. This title should live in the Lobby component since it is specific to that layout. For the onclick we can do a similar thing to the game card and it may be a good idea to add a boolean `disabled` prop (perhaps to GameCard) as well that would allow us to disable onclick events for that card.

src/app/_components/Lobby/SetUp/SetUp.tsx
```diff
@@ -2,13 +2,16 @@ import {
...
-	Button
+	Button,
+	Box, Divider,
 } from "@mui/material";
 import Chat from "@/app/_components/_sharedcomponents/Chat/Chat";
 import GameLinkCard from "../_subcomponents/GameLinkCard/GameLinkCard";
 import { useGame } from "@/app/_contexts/Game.context";
 import { ISetUpProps } from "../LobbyTypes";
 import { useRouter, useSearchParams } from "next/navigation"
+import React from "react";
```
danbastin, line 13: nit: move this back to the top for consistency

danbastin (approved):

PR #31, opened 2024-12-22
Che bato/game card icons

CheBato:
- Implementation of GameCards Power, HP, Damage, sentinel, shield, experience and upgrades.

### DIFF
--- a/src/app/_components/_sharedcomponents/Cards/GameCard/GameCard.tsx
+++ b/src/app/_components/_sharedcomponents/Cards/GameCard/GameCard.tsx
@@ -36,7 +36,6 @@ const GameCard: React.FC<IGameCardProps> = ({
 			sendGameMessage(["cardClicked", cardData.uuid]);
 		}
 	};
-
 	const handleClick = onClick ?? defaultClickFunction;
 	const cardBorderColor = (card: ICardData) => {
 		if (card.selected) return "yellow";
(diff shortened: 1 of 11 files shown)


src/app/_components/_sharedcomponents/Cards/GameCard/GameCard.tsx
```diff
@@ -143,6 +319,19 @@ const GameCard: React.FC<IGameCardProps> = ({
...
 			)}
 		</MuiCard>
+		{otherUpgradeCards.map((subcard, index) => (
+			<Box
+				key={subcard.uuid}
+				sx={{...styles.upgradeIconLayer,
+					backgroundImage: `url(${(cardUpgradebackground(subcard))})`,
+					bottom: `${index * 7 + 2}px`,
+					border: `1px solid ${subcard.selectable ? 'limegreen' : 'transparent'}`}}
+				onClick={() => upgradeClickFunction(subcard)}
```
danbastin, line 329: Let's comment out this onClick for now. I'm not sure that we want the upgrades to register click events here as most clicks on a card should be going to the attached cards. If we need to display options for clicking on an upgrade for targeting purposes, it is more likely we do that in a popup rather than on the board.

CheBato: sure thing! I was hoping we'd use the popup for this :D

danbastin (approved):

PR #39, opened 2025-01-03
Che bato/private lobbies

CheBato:
- Added ability to create private lobbies.
- Added link support for joining game.


danbastin:
I hit an error when trying to pick first player while I was player 2. Seems like the player name was not being set for the game. CardClicked should probably be switched to use the players id instead of their name, but it would probably be good to have it working either way.

CheBato:
> I hit an error when trying to pick first player while I was player 2. Seems like the player name was not being set for the game. CardClicked should probably be switched to use the players id instead of their name, but it would probably be good to have it working either way.

Absolutely seems like I missed this during my initial test through. I'll have a look and fix it up asap.

danbastin (approved):

PR #40, opened 2025-01-05
fix create game scrollarea

visualjerk:
Fixes the "Create Game" box to allow scrolling on small screens.

### DIFF
--- a/src/app/_components/HomePage/HomePagePlayMode.tsx
+++ b/src/app/_components/HomePage/HomePagePlayMode.tsx
@@ -70,8 +70,6 @@ const HomePagePlayMode: React.FC = () => {
     const styles = {
         wrapper: {
             height: '100%',
-            scrollbarColor: '#888888 rgba(0, 0, 0, 0)',
-            scrollbarWidth: 'thin',
         },
         tabStyles: {
             color: 'white',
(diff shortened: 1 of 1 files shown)


src/app/_components/HomePage/HomePagePlayMode.tsx
```diff
@@ -68,14 +68,19 @@ const HomePagePlayMode: React.FC = () => {
     }, []);
 
     const styles = {
+        wrapper: {
+            height: '100%',
+            scrollbarColor: '#888888 rgba(0, 0, 0, 0)',
+            scrollbarWidth: 'thin',
+        },
```
CheBato, line 75: Hey! Welcome aboard and thank you for helping out. One small comment, we will probably want to keep everything the same as possible including the scrollbars. Here is an example of the scroll bar we've been using and you can adjust the background color to fit. let me know if it looks alright with the wrapper and if you need anything else. 

```css
'::-webkit-scrollbar': {
    width: '0.2vw',
},
'::-webkit-scrollbar-thumb': {
    backgroundColor: '#D3D3D3B3',
    borderRadius: '1vw',
},
'::-webkit-scrollbar-button': {
    display: 'none',
},
transition: 'scrollbar-color 0.3s ease-in-out',
```

danbastin: We porbably want to go ahead and add this to our theme file so that we have this styling available site wide. For now I would just remove this scroll bar styling and we'll add in the centralized styling in another update.

visualjerk: Hi there and thanks for the fast review. I agree, it is probably best to have this style globally and remove any variations of it.

danbastin:
Looks good. If you can DM Cazargar on the Discord if you're in there we'll get you set up with the Contributor role tag so you can access the dev channels.

danbastin (approved):

PR #61, opened 2025-01-15
Che bato/end game screen

CheBato:
### GameBoard/page.tsx
- Added check that if lobbyState exists and a game doesn't exist that we send the user to the lobby.
- Added check if we have winners in the gameState. If we have winners we permanently open the preference page with the end game screen.

### Preferences/page.tsx
- Added Preferences screen for the homePage it is similar to the preferences currently it adds the block list and logout. 

### Controlhub.tsx
- Added Preferences link

### OpponentCardTray.tsx
- Changed sendManualDisconnectMessage to sendMessage since they do the same.

### SetUp.tsx (lobby)
- Added the manualDisconnect to the handleExit function since we know a user wishes to leave the lobby if they click exit.

### PreferencesComponent.tsx
- Redid the preference page so we receive the tabs, openState, title and subtitle from a parent component.
- Changed the background color so its more in line with the design
- minor design fixes for differentiating between homePage and gameBoard

### BlockListTab.tsx
- Added the BlockListTab to list of components for handling blocking users.

### Preferences.types.ts
- Added buttonFnc to preference buttons this way we send what the buttons does from a parent component
- remade the PreferenceProps to support receiving properties from a parent component

### CardSleevesTab.tsx
- Removed the scroll from cardSleevesTab and moved it to a parent component (this will make scrolls available in other screens).

### CurrentGameTab.tsx
- Added Concede functionality
- Minor design fix

### EndGameTab.tsx
- Added a new endGameTab which checks what type of game we are in and returns the appropriate component.

### GameOptionsTab.tsx & KeyboardShortcutsTab.tsx
- Minor design fix

### BlockedUser.tsx
- Added a new component that represents a row in blocked users accepts username for now.

### EndGameOptionsCustom.tsx
- New EndGameTab component for custom/private games

### EndGameOptionsQuickMatch
- New EndGameTab component for quickMatches
- I used the API call enter-queue to reenter the user into a queue. After multiple testing scenarios this was by far the most elegant and stable option.

### PreferenceButton.tsx
- Added buttonFunctionalities
- Added disabled check.

### VerticalTabs.tsx
- Added the scroll for overflow here so its accessible to all subTabs.
- Added the endGame screen tab, BlockedUserTab and Logout tab.

### Game.context.tsx
- Added resetStates which resets the states of GameState and LobbyState. I found that we need to do this to requeue a user into the queue otherwise the user will just be sent back into the game.
- Added socketKey as a userEffect dependable since when a user wanted to requeue we closed the connection with the socket but the user wasn't able to reconnect to a new socket since there wasn't a dependable triggered. This way when we change the socketKey for the user in resetStates we can make sure the user can requeue with a new socket.
- Removed sendManualDisconnectMessage since it was redundant (we can send the message via sendMessage()).

### User.context.tsx
- Added callbackUrl to logout function. Since if we logout in the preference page it redirects to the preference page which it shouldn't. This way we always redirect to the homepage.

### Lobby/page.tsx
- Changed how we check if a game is onGoing we have a new parameter in the lobbyState that gives us that information. We do this since otherwise when users want to rejoin the lobby from the match we can't send gameState as null from the BE hence why we set this parameter to allow users back into the lobby.

### quickGame/page.tsx
- minor design fixes.

-------------------------------------------------
### new changes
- Refactored the way time was count down when two users matched in the quickMatch
- Refactored both endgame screens to allow for users to send and receive rematch requests
- Refactored states so all rematches, requeues work correctly.
- Refactored requeue to use the same socket instead of an API call
- Fixes to certain hooks so we don't get warnings or errors when requeueing or rematching.


src/app/Preferences/page.tsx
```diff
@@ -0,0 +1,60 @@
...
+            fontSize: '0.90rem',
+        }
+    };
+
+    return (
+        <Box sx={styles.mainContainer}>
+            <Typography sx={styles.lobbyTextStyle} onClick={handleExit}>KARABAST</Typography>
+            <PreferencesComponent
+                isPreferenceOpen={true}
+                tabs={['keyboardShortcuts','cardSleeves','gameOptions','blockList','logout']}
```
danbastin, line 48: I think we can leave the logout out of the menu here since we have it in the top menu

src/app/_contexts/Game.context.tsx
```diff
@@ -36,6 +36,7 @@ export const GameProvider = ({ children }: { children: ReactNode }) => {
     const { openPopup, clearPopups } = usePopup();
     const { user } = useUser();
     const searchParams = useSearchParams();
+    const [socketKey, setSocketKey] = useState(0);
```
danbastin, line 39: What is this being used for?

danbastin (commented):
Couple of quick bugs:
- When returning to the lobby for a new game players should not be ready
- With the end game windows up after one player leaves the other play cannot close the window

src/app/GameBoard/page.tsx
```diff
@@ -40,7 +48,18 @@ const GameBoard = () => {
         if (drawerRef.current) {
             setDrawerWidth(drawerRef.current.offsetWidth);
         }
-    }, [sidebarOpen]);
+        if(lobbyState && !lobbyState.gameOngoing && lobbyState.gameType !== MatchType.Quick) {
+            router.push('/lobby');
+        }
+    }, [sidebarOpen, gameState, lobbyState]);
```
danbastin, line 54: missing router dep

danbastin, line 54: missing router dep

src/app/_components/_sharedcomponents/Preferences/_subComponents/VerticalTabs.tsx
```diff
@@ -92,27 +132,46 @@ function VerticalTabs({ tabs }:IVerticalTabsProps) {
...
-                    />
-                ))}
+                {tabs.map((tabName, idx) => {
+                    if (tabName === 'logout') {
+                        return (
+                            <Tab
+                                key={tabName}
+                                sx={styles.tab}
+                                label={renderLabels(tabName)}
+                                onClick={logout} // Your logout function here
```
danbastin, line 142: Let's clean up some of these comments that aren't providing code clarity.

danbastin (approved):

PR #67, opened 2025-01-26
Addressing the Base Damage and Deployed Leader styling.

DonovanDevelops:
Added a PNG red background how I saw other backgrounds being created. Tested both single and double digit numbers of base damage. Also adjusted the background for the leader card in the center display when the leader is deployed.

Double check my logic for determining if a Leader is deployed or not - I had an additional check for it on new line 116 that checked if it was in game, the variant was leader, and the card type was not leader. I believe that checking this way should also cover pilots from set 4 (which might be a leader upgrade?).

I could not locate a way to determine if the leader's or base's epic action had been used to add the styling for that exhaust icon in as well.


src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
```diff
@@ -37,6 +37,17 @@ const LeaderBaseCard: React.FC<ILeaderBaseCardProps> = ({
...
+        cardStyleDeployed: {
+            backgroundColor: 'transparent',
+            width: '10rem',
+            height: '7.18rem',
+            textAlign: 'center',
+            display: 'flex',
+            cursor: 'default',
+            m: '0em',
+            position: 'relative', // Needed for positioning the red box
+            border: '1px solid #FFFFFF',
```
danbastin, line 49: Lets add a little opacity to this border. I'll probably make a css var for this at some point since we'll want this to match that line elements as well, but for now lets just add some here to try and match the current unit bord borders.

src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
```diff
@@ -88,7 +113,7 @@ const LeaderBaseCard: React.FC<ILeaderBaseCardProps> = ({
     return (
         <Box>
             <Card
-                sx={isLobbyView ? styles.cardStyleLobby : styles.cardStyle}
+                sx={isLobbyView ? styles.cardStyleLobby : !isLobbyView && variant === 'leader' && card.type !== 'leader' ? styles.cardStyleDeployed : styles.cardStyle}
```
danbastin, line 116: We'll want to keep the nameplate visible when the leader is deployed.

public/damage-base.png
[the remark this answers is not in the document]
DonovanDevelops: Updated to be an SVG. I looked at trying to make it a background mask so we could _very easily_ adjust the color with backgroundColor, but unfortunately it tiles the SVG if it's not sized right for the scenario. Might be able to make it into a component somewhere for the future to let the one SVG be used for everything like sentinel, shields, exhasuted, and so on.

src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
DonovanDevelops: Updated the styling to match the unit arena borders.

src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
DonovanDevelops: All this change does is adjust the styling of the leader card (primary difference is the background image). This line doesn't imact the display of the namecard within the leader box.

danbastin (dismissed):

danbastin (approved):

PR #74, opened 2025-01-30
Added styling for cards that have not yet been implemented.

DonovanDevelops:
Added logic to the Card that determines if a card has been implemented or not and gives the user a visual. Currently works in-game.

Something to note - in GameCard.tsx, line 29 sets isFaceUp = true but nowhere else does it confirm that to build the card. When testing, the "not implemented" styling weas applying to the opponent's cards in hand, but I was not able to figure out why that was the case. I would have assumed that checking isFaceUp was the fix, so I am not certain if this is some residual thing.


DonovanDevelops:
image

danbastin (approved):

PR #103, opened 2025-02-16
Add lobby deck sort

Trys2Hard:
Lobby decks now sort by cost ascending. Added a slight hover effect to the "Exit Lobby" button.  Fixed a bug with set up component scrollbar not working.

### DIFF
--- a/src/app/_components/Lobby/SetUp/SetUp.tsx
+++ b/src/app/_components/Lobby/SetUp/SetUp.tsx
@@ -31,6 +31,8 @@ const SetUp: React.FC = ({
         sendMessage('manualDisconnect');
         router.push('/');
     }
+    // exit game lobby hover effect
+    const [isHovering, setIsHovering] = useState(false);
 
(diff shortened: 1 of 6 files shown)


src/app/_components/Lobby/SetUp/SetUp.tsx
```diff
@@ -55,13 +57,13 @@ const SetUp: React.FC = ({
...
             display: 'flex',
-            pr: '1.2em',
-            pl: '1.2em',
+            p: '1.2em',
             width: '100%',
             height: '10%',
             alignItems: 'center',
             justifyContent: 'space-between',
             cursor: 'pointer',
+            opacity: isHovering ? '0.8' : '1',
```
danbastin, line 66: Can we acheive this with just the CSS hover state? in this format we would add `'& :hover'` as a property under `exitCard`. Let me know if theres a reason this would not work.

danbastin (approved):

PR #114, opened 2025-02-18
Image Optimizations

dquilter:
Updates a number of images, reducing their file size:

- reduces image dimensions
- converts to a more appropriate formats (mostly to `webp`)

I noticed a couple of these images were stored on S3; I've moved them to the `public` folder for the purpose of getting a PR up, but am happy to move all of these updated images to S3 if we think that is the better place for them.


public/beta.jpg
danbastin, line 1: Lets do this one ins webp as well. Looks like a bit of quality loss on this one.

src/app/_components/_sharedcomponents/Banner/Banner.tsx
```diff
@@ -47,22 +47,22 @@ const KarabastBanner: React.FC = () => {
             flex: 1,
         },
         block1: {
-            backgroundImage: 'url(/leia-sor.jpg)',
```
danbastin, line 50: I've uploaded all of these as [name]-banner.webp for consistency

danbastin (approved):

PR #117, opened 2025-02-19
Add Preview Image to GameCard and LeaderBaseCard

dquilter:
Updates `LeaderBaseCard` and `GameCard` to include a preview on hover, using MUI Popover.

Handles a couple of instances where we might not want a preview:

- the Leader "box" if the leader has been deployed,
- the hidden cards in the opponents hand.


danbastin:
Really like the decision to use MUIPopover here. The animation is really nice. A few modifications if we can. 

1) Let's add a little bit of space between the card and the popover. I think 5px should be fine. Just enough for a little seperation. 
2) I think it would be good to add a small delay to this so the preview doesn't show as we move over elements. Lets start with 500ms and see how that feels. 
3) We'll probably want this for players cards in hand as well anchoring above the card. I could see a world where we put a longer delay on this, but lets start with the same delay and see how it is recieved.

danbastin:
Sorry, one more requirement addition I should have noted in my first pass. We'll want to also enable preview for subcards. This includes upgrades and capture card when those elements are hovered.

danbastin (approved):

PR #131, opened 2025-02-26
swudb regex link change

CheBato:
- Edited the deck regex to work as currentbast does.

### DIFF
--- a/src/app/api/swudbdeck/route.ts
+++ b/src/app/api/swudbdeck/route.ts
@@ -38,7 +38,6 @@ export async function GET(req: Request) {
             }
         }
         else if (deckLink.includes('swudb.com')) {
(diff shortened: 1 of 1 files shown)


src/app/api/swudbdeck/route.ts
```diff
@@ -38,7 +38,8 @@ export async function GET(req: Request) {
             }
         }
         else if (deckLink.includes('swudb.com')) {
-            const match = deckLink.match(/\/deck\/(?:view\/)?([^/?]+)/);
+            // const match = deckLink.match(/\/deck\/(?:view\/)?([^/?]+)/);
```
danbastin, line 41: If the new rule is more permissive and covers more cases I think we can just get rid of this line to keep the file clean.

PR #136, opened 2025-02-27
Update Dropdown Component with Autocomplete

dquilter:
Adds the MUI Autocomplete component to the DropdownPopup.


danbastin:
I think maybe the message isn't being sent to the server properly cause I can still play the named card when using Regional Governor. I can do some more investiagting tomorrow if needed.

dquilter:
> I think maybe the message isn't being sent to the server properly cause I can still play the named card when using Regional Governor. I can do some more investiagting tomorrow if needed.

I just pulled the latest server and now it does look a bit funky. Unlikely to get a chance to look till Monday, if you look before then, lemme know what you discover.

dquilter (commented):
Didn't quite appreciate how the index was changing during use, updated to work correctly.

danbastin (approved):

PR #163, opened 2025-03-02
- Added the ability to import json as a decklink

CheBato:
- Added additional validation steps so we don't have accept incorrect formats
- Changed that input fields now except texts
- Added json validation on FE so we know if the deck is correctly formated or not.
- Tested if the deck validation from BE correctly parses and responds.

### DIFF
--- a/src/app/_utils/checkJson.ts
+++ b/src/app/_utils/checkJson.ts
@@ -43,55 +43,58 @@ const isValidCardId = (id: string): boolean => {
 
 
 /**
- * Checks if an object only contains allowed keys
+ * Checks if an object only contains allowed keys (case-insensitive) and normalizes them
  * @param obj - Object to check
- * @param allowedKeys - Array of allowed key names
- * @returns boolean indicating if only allowed keys are present
+ * @param allowedKeys - Array of allowed key names in proper case format
+ * @returns An object containing a boolean indicating if only allowed keys are present and the normalized object
  */
-const hasOnlyAllowedKeys = (obj: Record<string, unknown>, allowedKeys: string[]): boolean => {
+const hasOnlyAllowedKeys = <T extends Record<string, unknown>>(
+    obj: T,
+    allowedKeys: string[]
+): { isValid: boolean; normalizedObj: Record<string, unknown> } => {
+    const normalizedObj: Record<string, unknown> = {};
+    const allowedKeysLower = allowedKeys.map(key => key.toLowerCase());
+
     for (const key of Object.keys(obj)) {
-        if (!allowedKeys.includes(key)) {
-            return false;
+        const keyLower = key.toLowerCase();
+        const allowedKeyIndex = allowedKeysLower.indexOf(keyLower);
+
+        if (allowedKeyIndex === -1) {
+            // Key not found in allowed keys (case-insensitive)
+            return { isValid: false, normalizedObj: {} };
         }
+
+        // Use the proper case from allowedKeys
+        const normalizedKey = allowedKeys[allowedKeyIndex];
+        normalizedObj[normalizedKey] = obj[key];
     }
-    return true;
+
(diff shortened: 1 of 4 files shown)


src/app/_utils/checkJson.ts
```diff
@@ -0,0 +1,103 @@
...
+ * @param jsonString - String to validate as JSON
+ * @returns Parsed DeckJSON object if valid, null if invalid
+ */
+export const validateDeckJSON = (jsonString: string): DeckJSON | null => {
+    try {
+        // Try to parse as JSON
+        const deckData = JSON.parse(jsonString);
+
+        // Validate the required fields
+        if (!deckData.metadata || !deckData.leader || !deckData.base || !deckData.deck) {
```
danbastin, line 42: Lets allso ad a check to ensure the deckIds are the `CCC_###` format that we get when using the link just to make sure it is correct but also that no one is pulling any funny business witht he ids.

src/app/_utils/checkJson.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        // Validate deck array
+        if (!Array.isArray(deckData.deck) || deckData.deck.length === 0) {
+            return null;
+        }
+
+        // Ensure all deck entries have id and count
+        for (const card of deckData.deck) {
+            if (!card.id || card.count === undefined) {
```
AMMayberry1, line 63: Need to use `==` here so it also catches `null`, same for sideboard check

src/app/_utils/checkJson.ts
```diff
@@ -0,0 +1,103 @@
...
+                return null;
+            }
+        }
+
+        // Validate sideboard if present
+        if (deckData.sideboard && Array.isArray(deckData.sideboard)) {
```
AMMayberry1, line 69: Should fail if sideboard exists but isn't an array

src/app/_utils/checkJson.ts
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Validates if a string is a properly formatted deck JSON
+ * @param jsonString - String to validate as JSON
+ * @returns Parsed DeckJSON object if valid, null if invalid
+ */
+export const validateDeckJSON = (jsonString: string): DeckJSON | null => {
+    try {
+        // Try to parse as JSON
+        const deckData = JSON.parse(jsonString);
```
AMMayberry1, line 39: Did some quick digging on sanitizing json inputs and it sounds like we probably don't have any major security risks since we're just reading values off the json. But it is recommended to add a maximum number of characters as the second parameter of `JSON.parse()` to avoid overflow attacks.

Also, a few places recommended filtering out unicode characters before parsing since those are more vulnerable to attacks.

src/app/_utils/checkJson.ts
AMMayberry1, line 42: One other type of check we should do is make sure that there aren't any unexpected keys at any layer

src/app/_utils/checkJson.ts
```diff
@@ -0,0 +1,103 @@
...
+        for (const card of deckData.deck) {
+            if (!card.id || card.count === undefined) {
+                return null;
+            }
+        }
+
+        // Validate sideboard if present
+        if (deckData.sideboard && Array.isArray(deckData.sideboard)) {
+            for (const card of deckData.sideboard) {
+                if (!card.id || card.count === undefined) {
```
AMMayberry1, line 71: As Caz mentioned, we also need to check that the values themselves are legal. So ID needs to be a number and count needs to be set-code-like (don't need to actually check that the set code is legal, the deck validator can do that. Just needs to be [AAA]_[111] format.) Probably could use a helper method for that since base and leader also need to be checked.

AMMayberry1 (approved):

danbastin (approved):

PR #164, opened 2025-03-02
Che bato/deck page

CheBato:
- Added Deck page for annonymous users.
- A player can add decks to localstorage from swustats or swudb.
- A player can add decks to localstorage via the favourites check on the main page.
- A player can add decks to localstorage in the lobby.
- A player can use decks from localstorage in the main menu for quick games and custom games
- A player can change his deck to one from localstorage in lobby.
- Added error handling when a favourite deck suddenly becomes unlisted or other error while still being in the localStorage.
- Added visual candy for the user when they take a look at their decks.
- Added 3 types of sorts in the deckPage. by favourites, name and deck source.
- Disabled order by in the detailed deckPage.


danbastin:
Really like how this page is shaping up. I have some nots but since a lot of this is on the fly and this is a relatively complex page it could serve as a good reason to catch up and chat. Let me know if you would have some time for that tomorrow. 

Some things I might look at if you were going to work on anything before we have a chance to talk:
1) I think for now we should just put the main deck and the sidebard into the same scroll box. Some people are reporting their deck getting cut off and until we have time to do something fancier I think it's better to just have them as once continue list. This applies to the lobby as well. 
2) one thing we can do to get this online quickly is just do what the current code does for now and save the link in local storage. This will be enough for decks and this page to be completely usable and we can add additional features from there if we decide to move things to the DB.

src/app/_components/_sharedcomponents/QuickGameForm/QuickGameForm.tsx
```diff
@@ -55,11 +67,25 @@ const QuickGameForm: React.FC<ICreateGameFormProps> = () => {
     const handleJoinGameQueue = async (event: FormEvent<HTMLFormElement>) => {
         event.preventDefault();
         setQueueState(true);
+        // Get the deck link - either from selected favorite or direct input
+        let userDeck = '';
+        console.log(favouriteDeck);
```
danbastin, line 72: Let's start to clean up some of these console logs where we can

src/app/_components/_sharedcomponents/CreateGameForm/CreateGameForm.tsx
```diff
@@ -185,24 +212,29 @@ const CreateGameForm = () => {
             </Typography>
             <form onSubmit={handleCreateGameSubmit}>
                 {/* Favourite Decks Input */}
-                {user && <FormControl fullWidth sx={styles.formControlStyle}>
+                <FormControl fullWidth sx={styles.formControlStyle}>
                     <Typography variant="body1" sx={styles.labelTextStyle}>Favourite Decks</Typography>
```
danbastin, line 216: Let's chanage this to "Favorite" in both places

src/app/_components/_sharedcomponents/CreateGameForm/CreateGameForm.tsx
```diff
@@ -262,7 +294,6 @@ const CreateGameForm = () => {
                         </Typography>
```
danbastin, line 294: Change these to "Save Deck List"

danbastin:
Overall really good stuff. I'm not gonna nit picky here cause the important thing is that it's functinoal and we can clean it up later. 

One thing I forgot to comment on is that I would really like for the lobby to default to showing the decks drop down at first if there are decks to be shown and have the button say "Load deck". Probably locaked above that would be a text link that says "import new deck" and clicking that would show the deck link input with a button saying "Import deck".

That may be a post feature thing unless you think you can whip that up quickly. But the purpose is just to save as much space there as we can but also be super clear about if the user is using one of their preexisting decks or loading up a new one. Let me know what you think.

CheBato:
> Overall really good stuff. I'm not gonna nit picky here cause the important thing is that it's functinoal and we can clean it up later.
> 
> One thing I forgot to comment on is that I would really like for the lobby to default to showing the decks drop down at first if there are decks to be shown and have the button say "Load deck". Probably locaked above that would be a text link that says "import new deck" and clicking that would show the deck link input with a button saying "Import deck".
> 
> That may be a post feature thing unless you think you can whip that up quickly. But the purpose is just to save as much space there as we can but also be super clear about if the user is using one of their preexisting decks or loading up a new one. Let me know what you think.

I got it working :D I would only kindly ask you if you could just move it to where it you think it would be best. Otherwise I think it was a great idea to include the swapping between one and the other. Let me know what you think!

Issue #156, opened 2025-03-01
Make "exit game" button clearer and give a warning before leaving


AMMayberry1:
This is causing confusion and it's too easy to accidentally drop the game

![Image](https://github.com/user-attachments/assets/afbd8b9e-c95f-4ccf-ae39-5b9baad5b841)

PR #166, opened 2025-03-02
#156 Popup when leaving the game

CheBato:
- Added popup for leaving the game.


danbastin:
We should check to see if the came is already ended and adjust teh message accordingly. just a simple "Leave the game and return to the home screen?"

CheBato:
> We should check to see if the came is already ended and adjust teh message accordingly. just a simple "Leave the game and return to the home screen?"

resolved! Thank you for that.

danbastin (approved):

PR #205, opened 2025-03-10
Move initiative token

Trys2Hard:
Initiative token moved to new location. It now slides off screen and back in when claimed.


src/app/_components/Gameboard/Board/Board.tsx
```diff
@@ -22,7 +23,17 @@ const Board: React.FC<IBoardProps> = ({
     const opponentLeader = gameState?.players[opponentId].leader;
     const opponentBase = gameState?.players[opponentId].base;
 
+    const hasInitiative = gameState.players[connectedPlayer].hasInitiative;
+    const initiativeClaimed = gameState.initiativeClaimed;
 
+    const slideIn = keyframes`
```
danbastin, line 29: This is unneeded if we're defining the animation as we are in the CSS. Mostly you would want to add keyframes if you animation is not linear.

src/app/_components/Gameboard/Board/Board.tsx
```diff
@@ -160,9 +273,16 @@ const Board: React.FC<IBoardProps> = ({
             <Grid container size="grow" sx={styles.ColumnStyle}>
                 <Box sx={styles.rightColumnBorderLeft} />
                 <Box sx={styles.rightColumnBorderRight} />
+                <Box sx={initiativeClaimed ? styles.opponentInitiativeClaimedWrapper : styles.opponentInitiativeWrapper }>
```
danbastin, line 276: I think we would want to absolutely position these relative to the units board since those are the corners we are putting this in. If they are absolutely positioned then we shouldn't need to change the height on those boards.

src/app/_components/Gameboard/Board/Board.tsx
```diff
@@ -125,6 +136,108 @@ const Board: React.FC<IBoardProps> = ({
             height: '100%',
             position: 'absolute',
         },
+        opponentInitiativeWrapper: {
+            transform: hasInitiative ? 'translate(100vw)' : 'translate(0vw)',
```
danbastin, line 140: This might require something done in the game context, but if we can get this animation to only fire if the initiative changed hands that would be ideal.

danbastin (approved):

PR #218, opened 2025-03-15
Discord and google authentication

CheBato:
To be able to set things up you do require discord and google secret keys and the NEXTAUTH_SECRET needs to be the same in the FE and BE.

DISCORD_CLIENT_ID
DISCORD_CLIENT_SECRET
NEXTAUTH_SECRET
GOOGLE_CLIENT_ID
GOOGLE_CLIENT_SECRET


src/app/_utils/DeckStorageUtils.ts
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * Retrieves a deck by its ID
+ * @param deckId Deck ID to retrieve
+ * @returns Promise that resolves to the deck data
+ */
+export const getDeckFromServer = async (deckId: string | string[]): Promise<IDeckDetailedData> => {
+    try {
+        // Make sure we have an anonymousUserId if needed
+
+        const response = await fetch(`${process.env.NEXT_PUBLIC_ROOT_URL}/api/get-deck/${deckId}`, {
```
AMMayberry1, line 273: Does this work if `deckId` is an array?

src/app/DeckPage/[DeckId]/page.tsx
```diff
@@ -71,75 +68,91 @@ const DeckDetails: React.FC = () => {
...
-                                }
-                                return;
-                            }
-                        }
+    const fetchDeckFromServer = async (deckId: string | string[]) => {
+        if (deckId) {
+            // we need to check if the deck is still available
+            try {
+                // we get the deck from localStorage and set the link
+                const deckDataServer = await getDeckFromServer(deckId);
```
Comment (AMMayberry1) [inline src/app/DeckPage/[DeckId]/page.tsx:123]: Left a comment on `getDeckFromServer` as well but it feels like there would need to be different handling here if `deckId` is an array

src/app/DeckPage/[DeckId]/page.tsx
```diff
@@ -71,75 +68,91 @@ const DeckDetails: React.FC = () => {
...
+                const data = await fetchDeckData(deckDataServer.deck.deckLink, false);
+                setDeckData(data);
+                setDisplayDeck(deckDataServer);
+                calculateStats(deckDataServer.stats);
+                getFormattedOpponentStats(deckDataServer.stats?.statsByMatchup);
+            } catch (error) {
+                // check if its set to private
+                if (error instanceof Error) {
```
Comment (AMMayberry1) [inline src/app/DeckPage/[DeckId]/page.tsx:131]: When would this not be the case?

src/app/DeckPage/page.tsx
```diff
@@ -31,13 +29,34 @@ const DeckPage: React.FC = () => {
     const [addDeckDialogOpen, setAddDeckDialogOpen] = useState<boolean>(false);
     const [selectedDecks, setSelectedDecks] = useState<string[]>([]); // Track selected decks
     const [deleteDialogOpen, setDeleteDialogOpen] = useState<boolean>(false);
+    const [ConfirmationDialogMessage, setConfirmationDialogMessage] = useState<string>('');
     const router = useRouter();
+    const { user } = useUser();
 
     // Load decks from localStorage on component mount
     useEffect(() => {
```
AMMayberry1, line 37: This feels like a bit of a hack, is there not an explicit way to do this in React?

src/app/DeckPage/page.tsx
```diff
@@ -31,13 +29,34 @@ const DeckPage: React.FC = () => {
...
     }, []);
 
+    // Function to load decks from localStorage and server
+    const fetchDecks = async () => {
+        try {
+            // Call the loadDecks function and await the result
+            const fetchedDecks = user ? await loadDecks() : await loadSavedDecks();
+            // Update state with the fetched decks converted to display format
+            // Default to favourites first if sort option is unrecognized
+            fetchedDecks.sort((a, b) => {
```
AMMayberry1, line 48: This can be a TODO for a future PR but as a second layer of sorting it might be good to prioritize recently used decks. Could mean played in a game, and if swudb publishes a "last update" time we could also use that

src/app/DeckPage/page.tsx
```diff
@@ -139,48 +154,54 @@ const DeckPage: React.FC = () => {
...
-
+        try{
+            await deleteDecks(selectedDecks);
+            // Update deck list in state
+            setDecks(prevDecks => prevDecks.filter(deck => !selectedDecks.includes(deck.deckID)));
+            // Reset selection
+            setSelectedDecks([]);
+        } catch (error){
+            console.log(error);
+            setConfirmationDialogMessage('There was an error when deleting decks. Try again later. '+error);
```
AMMayberry1, line 203: nit: might be better as "Please try again later"

src/app/_components/HomePage/PublicGames/PublicGames.tsx
```diff
@@ -7,7 +7,7 @@ import { SwuGameFormat } from '@/app/_constants/constants';
 
 const fetchLobbies = async (setLobbies: (lobbies: ILobby[]) => void) => {
     try {
-        const response = await fetch(`${process.env.NEXT_PUBLIC_ROOT_URL}/api/available-lobbies`);
+        const response = await fetch(`${process.env.NEXT_PUBLIC_ROOT_URL}/api/available-lobbies`,{ credentials:'include' });
```
AMMayberry1, line 10: Does this `credentials` thing get used in the API call?

src/app/_components/_sharedcomponents/QuickGameForm/QuickGameForm.tsx
```diff
@@ -45,17 +45,27 @@ const QuickGameForm: React.FC<ICreateGameFormProps> = () => {
...
+        fetchDecks();
+    }, [user]);
 
 
     // Load saved decks from localStorage
-    const loadDecks = () => {
-        const decks = loadSavedDecks();
-        if(decks.length > 0) {
-            setFavouriteDeck(decks[0].deckID);
+    const fetchDecks = async () => {
```
AMMayberry1, line 53: Definitely should consolidate this to a method, it's repeated several times

src/app/_contexts/User.context.tsx
```diff
@@ -37,28 +38,51 @@ export const UserProvider: React.FC<{ children: ReactNode }> = ({
...
+                            email: session.user.email || null,
+                            provider: session.user.provider || null,
+                            providerId: session.user.id || null,
+                        }
+                    });
+                } catch (error) {
+                    console.error('Error syncing user with server:', error);
+                    alert('Server error while logging in');
+                    // You could set some error state here, or fall back to using just session data
+                    if(!storedUser){
```
AMMayberry1, line 60: This logic is too complex to be living inside of a catch block, there are ways it could conceivably throw another exception and that's real bad when you're inside an exception handler. Is there a way we could trigger it outside of the block?

src/app/_utils/DeckStorageUtils.ts
```diff
@@ -0,0 +1,289 @@
...
+        });
+        const returnedData = await response.json();
+        if (!response.ok) {
+            const error = await response.json();
+            console.error('Error saving deck to server:', error);
+            throw new Error('Error when attempting to save deck. '+ error);
+        }
+        return returnedData.deck.id;
+    } catch (error) {
+        throw error;
```
AMMayberry1, line 100: We might need to be a little more careful with throwing in the FE since that will just break the screen completely. We can try being really careful to have try-catch handlers around all of these calls but I wonder if it's better for this side of things to just log and error and return `null` or something that makes it obvious that there was a failure.

Might be worth discussing with Caz and / or the other FE folks, I'm less familiar with best practices for error handling in React.

src/app/_utils/s3Utils.ts
```diff
@@ -23,6 +23,7 @@ export const s3CardImageURL = (card: ICardData | IServerCardData | ISetCode | IP
 
     // we check which type it is
     const isGameOrSetCard = isGameCard(card) || isSetCodeCard(card) || isPreviewCard(card);
+    console.log(card.id);
```
AMMayberry1, line 26: Was this for debugging purposes? I feel like it will generate a lot of noise, if we want to keep it then maybe we do it only in development mode

src/app/DeckPage/page.tsx
CheBato: Absolutely, added a github issue so we don't forget

src/app/_components/_sharedcomponents/QuickGameForm/QuickGameForm.tsx
CheBato: this is as much as i can consolidate.

src/app/DeckPage/page.tsx
CheBato: I can see how this might look a bit hacky but its a standard React pattern for fetching data using React hooks.

src/app/_components/HomePage/PublicGames/PublicGames.tsx
CheBato: yes it does.

CheBato: not in this one i removed the credentials from the calls that don't use authentication.

src/app/_utils/DeckStorageUtils.ts
CheBato: there are 4 situations where we use saveDeck to server and in two of these situations I can see this being being okay to return null. Those are the CreateGameForm and QuickGameForm. In the SetupCard we display an error to the user to let him know that his deck was not saved to favorites and we do the same in the deckPage when a user adds a deck.

src/app/DeckPage/[DeckId]/page.tsx
Comment (CheBato) [reply src/app/DeckPage/[DeckId]/page.tsx:131]: We need to check since by default the error is of type unknown.

AMMayberry1:
General note that we should probably show the username somewhere on the home screen to help it make it clearer that they're logged in. In the general vicinity of the log-out button would probably be good, but it might be a good idea to ask the Design crew for some ideas on how to clean up that whole top-right region of the screen. it's getting pretty busy

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GeneralTab.tsx
```diff
@@ -0,0 +1,143 @@
+import * as React from 'react';
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
+import { Button, Divider, Link, TextField } from '@mui/material';
+import { useUser } from '@/app/_contexts/User.context';
+import { useEffect, useState } from 'react';
+import { ErrorModal } from '@/app/_components/_sharedcomponents/Error/ErrorModal';
+import { setUsernameOnServer } from '@/app/_utils/DeckStorageUtils';
+
+function GeneralTab() {
```
AMMayberry1, line 10: Just a note, the Preferences tab is missing a Back button

AMMayberry1, line 10: Leaving a random comment here to track that we need to remove the Profile button since we have Preferences (or swap them, either way)

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GeneralTab.tsx
```diff
@@ -0,0 +1,143 @@
...
+                        <TextField
+                            value={username}
+                            onChange={(e) => {
+                                setUsername(e.target.value);
+                                setDeckErrorSummary(null);
+                                setDeckErrorDetails(undefined);
+                            }}
+                            sx={styles.textFieldStyle}
+                        />
+                        <Button variant="contained" onClick={handleChangeUsername} sx={styles.buttonStyle}>
```
AMMayberry1, line 118: Would be good to disable the "change username" button until a new value is typed. I just unconsciously pressed it thinking it would open a new page to do the change, which triggered "changing" my username to the value it already is.

Honestly wouldn't be a bad idea to have a filter in at least one spot that confirms that the old and new username are different - since we're rate-limiting this we want to be real careful to prevent accidents so they don't use up their allotted name changes and we get complaints about it

AMMayberry1, line 118: This is probably a more appropriate comment for the BE but it would be good to print out the timestamp when the user will be allowed to change their username next so they can know

src/app/_components/_sharedcomponents/DeckPage/AddDeckDialog.tsx
```diff
@@ -54,6 +59,11 @@ const AddDeckDialog: React.FC<AddDeckDialogProps> = ({
                     setDeckErrorSummary('Couldn\'t import. Deck is invalid.');
                     setErrorModalOpen(true);
                 }
+            }else{
+                setErrorTitle('Server error');
+                setDeckErrorSummary('Server error when saving deck to server.');
```
AMMayberry1, line 64: General note for this whole PR and feature - I think every error message related to user data should include the user's ID so we can look up their data in the DB.

I have something like this for the quick match search window already, but probably it and all the others should include the ID but also say "please save this value and send it to us in the discord."

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GeneralTab.tsx
AMMayberry1, line 118: There needs to be some kind of visual feedback on a successful name change

src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -1,8 +1,11 @@
 import NextAuth from 'next-auth';
 import GoogleProvider from 'next-auth/providers/google';
 import DiscordProvider from 'next-auth/providers/discord';
+import type { JWT } from 'next-auth/jwt';
+import jwt from 'jsonwebtoken';
```
Comment (AMMayberry1) [inline src/app/api/auth/[...nextauth]/route.ts:5]: Not sure where to put this so leaving it here - we need to hide the "Order66" and "ThisIsTheWay" logins when we're not in dev mode

src/app/_contexts/Game.context.tsx
```diff
@@ -56,8 +57,12 @@ export const GameProvider = ({ children }: { children: ReactNode }) => {
     const searchParams = useSearchParams();
     const router = useRouter();
     const [distributionPromptData, setDistributionPromptData] = useState<IDistributionPromptData | null>(null);
-
+    const { data: session, status } = useSession();
```
AMMayberry1, line 60: Another general note (not related to this code specifically) - we'll need to set up filtering for spectators so that they don't see usernames in the chat window. I kicked off a discussion in dev-general on what names to use, but we can just do whatever's simplest to start (probably just use "Player 1" and "Player 2")

src/app/DeckPage/[DeckId]/page.tsx
```diff
@@ -134,9 +134,12 @@ const DeckDetails: React.FC = () => {
                         setDeckErrorDetails({
                             [DeckValidationFailureReason.DeckSetToPrivate]: true,
                         });
+                    }else if(error.message.includes('Authentication error')){
+
```
Comment (AMMayberry1) [inline src/app/DeckPage/[DeckId]/page.tsx:138]: Missing code?

src/app/DeckPage/[DeckId]/page.tsx
```diff
@@ -352,58 +369,41 @@ const DeckDetails: React.FC = () => {
...
-                            <Box>
-                                <Box sx={styles.gamesRow}>
-                                    <Box sx={{ ...styles.gamesCircle, backgroundColor:'#367684' }} />
-                                    <Typography variant={'h5'} sx={{ color: '#fff' }}>Games Played: 55</Typography>
-                                </Box>
-                                <Box sx={styles.gamesRow}>
-                                    <Box sx={{ ...styles.gamesCircle, backgroundColor:'#6CF3D3' }} />
-                                    <Typography variant={'h5'}>Games Won: 30</Typography>
+                    {/* Stats go here */}
+                    {displayDeck && (
```
Comment (AMMayberry1) [inline src/app/DeckPage/[DeckId]/page.tsx:376]: Let's hide stats completely when not logged in, maybe even leave a note explaining that stats are only available to logged-in users

src/app/DeckPage/page.tsx
```diff
@@ -82,11 +93,9 @@ const DeckPage: React.FC = () => {
...
-    const loadDecks = () => {
-        const decks = loadSavedDecks();
-        setDecks(convertToDisplayDecks(decks));
-    }
+    const handleBackButton = () => {
+        router.push('/');
+    };
 
     // Handle successful deck addition
     const handleAddDeckSuccess = (deckData: IDeckData, deckLink: string) => {
```
AMMayberry1, line 101: Looks like using the "add deck" button to add a deck link that already exists will wipe out the stats for that deck. Definitely this should be a no-op, maybe pop up a message saying "deck already exists" but that's optional for now

src/app/api/auth/[...nextauth]/route.ts
Comment (CheBato) [reply src/app/api/auth/[...nextauth]/route.ts:5]: It is already implemented in Login.tsx

CheBato:
I need to add a general reminder comment that we need to handle swu-unlimited decks in the deck page currently we do get the decks from there, but they aren't tagged correctly.

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GeneralTab.tsx
```diff
@@ -0,0 +1,155 @@
...
+            color: 'var(--selection-red);',
+            textDecorationColor: 'var(--initiative-red);',
+        }
+    }
+
+    return (
+        <>
+            <Box sx={styles.functionContainer}>
+                <Typography sx={styles.typographyContainer} variant={'h2'}>Profile</Typography>
+                <Divider sx={{ mb: '20px' }}/>
```
AMMayberry1, line 113: I can't remember if I said it already, but we should consider adding a text box with the user ID so that we can ask people to send it to us for checking the logs or DB for debugging purposes, ideally for both logged-out and logged-in users. Otherwise I'm not sure how we'll be able to do debugging at all when users hit issues related to logins

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GeneralTab.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        <TextField
+                            value={username}
+                            onChange={(e) => {
+                                setUsername(e.target.value);
+                                setDeckErrorSummary(null);
+                                setDeckErrorDetails(undefined);
+                            }}
+                            sx={styles.textFieldStyle}
+                        />
+                        <Button variant="contained" disabled={username.length === 0 || !(user?.username != username)} onClick={handleChangeUsername} sx={styles.buttonStyle}>
```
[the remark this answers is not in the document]
AMMayberry1: Couple other issues:
- I wiped my local db and created a new user, the button was grayed out even though it's within the 1hr limit
- Thinking about it now, we definitely should have a text box above the username explaining the rules for renaming frequency (1hr then 1 month or w/ever it is)

AMMayberry1:
I can't consistently reproduce this, but a handful of times when I log in, add a deck, then log out, I see a popup like "error fetching decks" and see this in the logs:

image

Looks like it might be trying to fetch decks from the DB for an unauthenticated user? I can't yet give a consistent repro, lmk if you're not sure where to look for this and I can keep trying

AMMayberry1:
The login page doesn't have a back button, just realized

AMMayberry1:
This can be a later thing if it's hard to do, but definitely the tags for the different deckbuilders should be in different colors

src/app/api/auth/[...nextauth]/route.ts
Comment (AMMayberry1) [reply src/app/api/auth/[...nextauth]/route.ts:5]: Closing this out - with the added context from Caz I think NODE_ENV is fine, though we may eventually want to add some comments / wiki text which explains how to test in prod mode

src/app/_components/_sharedcomponents/ControlHub/ControlHub.tsx
```diff
@@ -89,14 +88,11 @@ const ControlHub: React.FC<IControlHubProps> = ({
                     <NextLinkMui href="/DeckPage" sx={styles.profileLink}>
                         Decks
                     </NextLinkMui>
+                    <NextLinkMui href="/Preferences" sx={styles.profileLink}>
+                        Preferences
+                    </NextLinkMui>
```
AMMayberry1, line 93: Adding the comment here - the "Login" button is not enabled in Prod mode (line 107). you can do the following to test prod mode:

1. copy-paste .env.development as .env
2. add a .gitignore entry for .env  **(IMPORTANT, SEE BELOW)**
3. run `npx next build`
4. run `npm start`

For #2 above, right now the .env file you add will show as tracked in git and we need to be _very_ careful that the file with secrets is never committed, even temporarily

src/app/page.tsx
```diff
@@ -32,7 +31,8 @@ const Home: React.FC = () => {
             fontSize: '0.75rem',
         },
     };
-
+    // i need to fix sync and add functionality to the popup for changing usernames.
```
AMMayberry1, line 34: What's this about?

CheBato: old comment i left

CheBato: old comment i left unfortunatelly when i was trying to find a place where to place the popup

CheBato: old comment i left unfortunatelly when i was trying to find a place where to place the popup

CheBato: old comment i left unfortunatelly when i was trying to find a place where to place the popup

src/app/_components/Auth/Login/Login.tsx
```diff
@@ -79,6 +81,13 @@ const Login: React.FC = () => {
 						    >
     Login as ThisIsTheWay
 						    </Button>
+						    <Button
+						        variant="contained"
+						        sx={{ ...loginStyles.button }}
+						        onClick={() => router.push('/')}
+						    >
+    Back to homepage
```
AMMayberry1, line 89: Can just say "Back" or else use the little arrow button we've used on other screens

src/app/_components/_sharedcomponents/HomescreenWelcome/WelcomePopup.tsx
```diff
@@ -0,0 +1,216 @@
...
+                    Before you start playing, we recommend setting a username that other players will see during games.
+                </Typography>
+
+                <Typography variant="body1" sx={styles.message}>
+                    Would you like to set your username now?
+                </Typography>
+
+                <Box sx={styles.textFieldContainer}>
+                    <TextField
+                        placeholder="Your Username"
```
AMMayberry1, line 167: We should probably set the placeholder to whatever their current name is so they know what they'll get if they don't change right now

src/app/_components/_sharedcomponents/ControlHub/ControlHub.tsx
CheBato: i got this running.  But it would seem I'm not able to communicate with the server for some reason. Could be something wrong with my configuration.

AMMayberry1 (approved):

PR #223, opened 2025-03-16
Fix Trench leader unit image

Trys2Hard:
Leader units deployed now have an image even if deployed without using an epic action.


src/app/_utils/s3Utils.ts
```diff
@@ -32,7 +32,7 @@ export const s3CardImageURL = (card: ICardData | IServerCardData | ISetCode | IP
 
     let cardNumber = setId.number.toString().padStart(3, '0')
 
-    if (isGameCard(card) && ((cardType === 'leaderUnit' && card.epicDeployActionSpent) || (cardType === 'leaderUpgrade'))) {
```
danbastin, line 35: This was added to differentiate actual leader units from vehicles that would become leaderUnits by having leaders attached to them. We're gonna need to figure out a way to determine if a unit is actually a leader or not. 

Maybe what we should do is make undeployed leaders be the ones with the extra label and that way we can check those zone to figure out if we need to render the landscape view rather than having the portrait be the exception.

PR #263, opened 2025-04-01
Sw unlimited db support

Beregost:


src/app/api/swudbdeck/route.ts
```diff
@@ -60,6 +60,30 @@ export async function GET(req: Request) {
...
+            deckSource = DeckSource.SWUnlimitedDB;
+            if (!deckId) {
+                console.error('Error: Invalid deckLink format');
+                return NextResponse.json(
+                    { error: 'Invalid deckLink format' },
+                    { status: 400 }
+                );
+            }
+
+            const apiUrl = `https://sw-unlimited-db.com/umbraco/api/deckapi/get?id=${deckId}`;
```
AMMayberry1, line 77: Just curious, what does `umbraco` in the path here refer to?

src/app/api/swudbdeck/route.ts
```diff
@@ -60,6 +60,30 @@ export async function GET(req: Request) {
...
+
+            response = await fetch(apiUrl, { method: 'GET' });
+            if (!response.ok) {
+                if(response.status === 404) {
```
AMMayberry1, line 81: Just to confirm - it's 404 whether the deck doesn't exist or it's just not published?

src/app/api/swudbdeck/route.ts
Beregost: it sounds like it would be, but no - umbraco is a website CMS, kind of like wordpress. I'm guessing they used it to build their site, which is why it would be in their api path.

This particular url I grabbed from the petranaki github: https://github.com/SWU-Petranaki/SWUOnline/blob/ec701d11b8e0112cf299dbe6a7ee18886301eb5a/JoinGameInput.php#L158

AMMayberry1 (approved):

PR #270, opened 2025-04-04
Add support for Force icon and re-enable Next Set Preview

AMMayberry1:
Added the Force icon created by Matavatar and did an integration test with the new property on the BE.

Update the image download scripts to point to LOF.

For some reason nextjs lint was hitting a bunch of stuff it never card about before, disabled linting for now until we can get vscode auto-lint to support it.


brianries (approved):

moyerr (approved):
I can't comment on the code, but I tested with Qui-Gon Gin and Darth Maul, along with the new bases. Everything looks good! Force Token appears/disappears when it should, and the color is correct for heroism/villainy aspects. Great work!

danbastin (dismissed):

brianries (approved):

PR #283, opened 2025-04-08
Che bato/bug reporting fe

CheBato:
The BE PR has instructions how to set up the webhook on discord.
BE Pull request: https://github.com/SWU-Karabast/forceteki/pull/1057

There could be some positioning or perhaps styling changes.

### DIFF
--- a/src/app/_components/Gameboard/OpponentCardTray/OpponentCardTray.tsx
+++ b/src/app/_components/Gameboard/OpponentCardTray/OpponentCardTray.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import { CloseOutlined, SettingsOutlined } from '@mui/icons-material';
+import React, { useState } from 'react';
+import { CloseOutlined, SettingsOutlined, BugReportOutlined } from '@mui/icons-material';
 import { Box, Grid2 as Grid, Popover, PopoverOrigin } from '@mui/material';
 import Resources from '../_subcomponents/PlayerTray/Resources';
 import PlayerHand from '../_subcomponents/PlayerTray/PlayerHand';
@@ -11,10 +11,12 @@ import { v4 as uuidv4 } from 'uuid';
 import { usePopup } from '@/app/_contexts/Popup.context';
 import { PopupSource } from '@/app/_components/_sharedcomponents/Popup/Popup.types';
 import { useRouter } from 'next/navigation';
+import BugReportDialog from '@/app/_components/_sharedcomponents/Preferences/_subComponents/BugReportDialog';
 
 const OpponentCardTray: React.FC<IOpponentCardTrayProps> = ({ trayPlayer, preferenceToggle }) => {
     const { gameState, connectedPlayer, getOpponent, isSpectator } = useGame();
     const { openPopup } = usePopup();
+    const [bugReportOpen, setBugReportOpen] = useState<boolean>(false);
     const router = useRouter();
     const handleExitButton = () => {
         if (isSpectator){
@@ -47,6 +49,16 @@ const OpponentCardTray: React.FC<IOpponentCardTrayProps> = ({ trayPlayer, prefer
     const handlePreviewClose = () => {
         clearTimeout(hoverTimeout.current);
         setAnchorElement(null);
+    };
+
+    // Handler for opening the bug report dialog
+    const handleOpenBugReport = () => {
+        setBugReportOpen(true);
+    };
+
+    // Handler for closing the bug report dialog
+    const handleCloseBugReport = () => {
+        setBugReportOpen(false);
     };
 
     const popoverConfig = (): { anchorOrigin: PopoverOrigin, transformOrigin: PopoverOrigin } => {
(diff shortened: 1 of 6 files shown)


src/app/_components/_sharedcomponents/Preferences/_subComponents/BugReportDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+            <StyledDialogActions>
+                <PreferenceButton
+                    buttonFnc={handleClose}
+                    variant="concede"
+                    text={'Cancel'}
+                />
+                <PreferenceButton
+                    buttonFnc={handleSubmit}
+                    variant="standard"
+                    disabled={!description.trim() || description.length > MAX_DESCRIPTION_LENGTH || success}
```
AMMayberry1, line 181: It's a little awkward that the user has to cancel out of the dialog after submitting. We should do something like replace the button text to say "Close" or else auto-close and pop up a message or something

src/app/_components/_sharedcomponents/Preferences/_subComponents/BugReportDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+                        <StyledTextField
+                            label="Bug Description"
+                            multiline
+                            rows={5}
+                            fullWidth
+                            variant="outlined"
+                            value={description}
+                            onChange={handleDescriptionChange}
+                            error={!!error}
+                            helperText={error || `${description.length}/${MAX_DESCRIPTION_LENGTH} characters`}
```
AMMayberry1, line 166: Does MUI have a convenient way to enforce a character limit? I feel like that might come standard. If not then this is fine, not worth writing our own method for it (though maybe we could turn this text red when the limit is exceeded)

src/app/_components/_sharedcomponents/Preferences/_subComponents/BugReportDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+            setDescription('');
+            setError('');
+            setSuccess(false);
+        }, 300);
+    };
+    return (
+        <StyledDialog open={open} onClose={handleClose}>
+            <StyledDialogTitle>Report a Bug</StyledDialogTitle>
+            <StyledDialogContent>
+                {(success || connectedUser?.reportedBugs >= 3) ? (
```
AMMayberry1, line 147: The UI should give clearer feedback when the max number of reported bugs is exceeded, I think users would interpret the current behavior as a bug (ironically)

src/app/_components/_sharedcomponents/Preferences/_subComponents/BugReportDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+        // Reset the form
+        setTimeout(() => {
+            setDescription('');
+            setError('');
+            setSuccess(false);
+        }, 300);
+    };
+    return (
+        <StyledDialog open={open} onClose={handleClose}>
+            <StyledDialogTitle>Report a Bug</StyledDialogTitle>
```
AMMayberry1, line 145: Still wondering if it would be good to have this button appear underneath the gear with like a bug symbol like you suggested? I could go either way, but I like the idea of making it more visible

src/app/_components/_sharedcomponents/Preferences/_subComponents/BugReportDialog.tsx
brianries: yeah agreed, I would turn the button green with "Ok" or "Done"

src/app/_components/_sharedcomponents/Preferences/_subComponents/BugReportDialog.tsx
CheBato: it unfortunatelly does not.

src/app/_components/_sharedcomponents/Preferences/_subComponents/BugReportDialog.tsx
CheBato: added a bug report icon!

src/app/_components/Gameboard/OpponentCardTray/OpponentCardTray.tsx
```diff
@@ -187,9 +199,14 @@ const OpponentCardTray: React.FC<IOpponentCardTrayProps> = ({ trayPlayer, prefer
                 <Box sx={styles.menuStyles}>
                     <CloseOutlined onClick={handleExitButton} sx={{ cursor:'pointer' }}/>
                     <SettingsOutlined onClick={preferenceToggle} sx={{ cursor:'pointer' }} />
+                    <BugReportOutlined onClick={handleOpenBugReport} sx={{ cursor:'pointer' }} />
                 </Box>
-
             </Grid>
+            {/* Bug Report Dialog */}
+            <BugReportDialog
```
AMMayberry1, line 206: I just realized that the spectator view has access to the bug report dialog currently. We should either remove it outright _or_ make it so that it pops up a similar modal as the `numReports > 3` case where it just says something like "spectators aren't allowed to report, feel free to report at our discord (link)".

It's such an edge case I don't think it will really matter what we do, so it's up to your preference

AMMayberry1 (approved):

PR #293, opened 2025-04-12
Preview on hover v2

Beregost:


Beregost:
firefox_eG4kjmFIdj
firefox_3SPA5EAbRf
firefox_7ohA5esgg7

src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
```diff
@@ -29,7 +29,6 @@ const LeaderBaseCard: React.FC<ILeaderBaseCardProps> = ({
 
     const handlePreviewOpen = (event: React.MouseEvent<HTMLElement>) => {
```
danbastin, line 30: Not entirely sure if this is related, but perhaps you could take a look at this case since we're making changes here. 

image

The hover previw is being shown when the leader is deployed.

Beregost: what would you like the behavior to be instead? Do you want to show an on-hover of the landscape side of the leader card?

Beregost: firefox_H6wBxrZnpK
fixed!

AMMayberry1:
Similar comment to Caz, found one other broken hover in another page - FoundGame in this case. If this would be a quick fix it'd be great to ahead and roll it into this PR:

image

src/app/_components/HomePage/_subcomponents/JoinableGame/JoinableGame.tsx
```diff
@@ -76,6 +111,13 @@ const JoinableGame: React.FC<IJoinableGameProps> = ({ lobby }) => {
             boxSizing: 'border-box',
             cursor: 'pointer'
         },
+        cardPopover: {
+            borderRadius: '.38em',
+            backgroundSize: 'cover',
+            backgroundRepeat: 'no-repeat',
+            aspectRatio: '1.4 / 1',
+            width: '16rem',
```
AMMayberry1, line 119: Not sure if this was discussed already, but I'd be in favor of making the hover images in the main menu a smidge bigger - it's still pretty hard to read the text on them

Beregost: fixed

AMMayberry1 (approved):

PR #296, opened 2025-04-13
Board and UnitBoard layout changes

brianries:
The scope of this project was to make scaling changes to the UnitBoard and the Border Edges around the board.  The goal was to enforce some more strict tiling / sizing of the game cards given the various resolutions.  We want to prevent units crossing the center line *for almost all cases* but it will still cross the center on small resolution and very large unit counts (8+).  I also created a debug class - BreakPointOverlay so you can enable it and get visual feedback on which 'responsive threshold' you're operating with

A few additional things had to be fixed after this rescaling:
- I added a 3rd segment to the border outlines that can dynamically scale (it's just a horizontal line that fills inbetween the left and right side).  I scaled down the pixel width of the small curve, and let the horizontal line fill the rest.
- I fixed the upgrade icon / text on smaller images.  You can see in the first image that it was rendering more box then it needed.  This is still not perfect and should probably get redesigned, but it works.

Here is a good before and after, using the iPadAir resolution:

iPadAir before:
image

iPadAir after:
image

iPhone12 before:
image

iPhone12 after:  (Still needs work, but closer!)
image

Example of the border separating on wider monitors (this is now fixed by the 3rd segment that fills the gap)
image


danbastin:
I like this as a start. One think to look at is how damaged and "mid-card" tokens like sentinel get rendered. 

image

This is using devtools iPad Air resolution. I think for something like this we could go a little bigger with the card and a little smaller with the tokens. 

Probably something we want to look into as well is stacking token units on smaller screens. Things like a crazy Endless Legion are always going to be tough but I think thats enough of an edgecase that would could have something like scrolling for the unit containers that is a less graceful handling but fine due to the rarity of the situation. Token units are a more likely situation so stacking toekns that are all the same stats could be an option.

AMMayberry1:
Looks like the "default" size for everything got smaller, even on a mostly-empty board and a PC screen (see comparison below). I might just be old and going blind but I liked things being a bit bigger when possible - would it be doable to get back to that without messing up the scaling on smaller screens? Also good to gauge whether people other than me would like that, I guess.

image

image

brianries:
> I like this as a start. One think to look at is how damaged and "mid-card" tokens like sentinel get rendered.
> 
> image
> 
> This is using devtools iPad Air resolution. I think for something like this we could go a little bigger with the card and a little smaller with the tokens.
> 
> Probably something we want to look into as well is stacking token units on smaller screens. Things like a crazy Endless Legion are always going to be tough but I think thats enough of an edgecase that would could have something like scrolling for the unit containers that is a less graceful handling but fine due to the rarity of the situation. Token units are a more likely situation so stacking toekns that are all the same stats could be an option.

Updated the sizing of the tokens and converted to "em" and percentage space.  Upped the size of the cards a tiny bit at the ipadAir resolution level.

brianries:
> Looks like the "default" size for everything got smaller, even on a mostly-empty board and a PC screen (see comparison below). I might just be old and going blind but I liked things being a bit bigger when possible - would it be doable to get back to that without messing up the scaling on smaller screens? Also good to gauge whether people other than me would like that, I guess.
> 
 
 Upped the size of the game cards to 7rem for this setting (don't shrink them just yet)

AMMayberry1:
Guessing you didn't mean to leave this enabled (though it was very helpful for the test):

image

AMMayberry1 (approved):

PR #309, opened 2025-05-02
Chat hover v3

Beregost:


Beregost:
firefox_WTYXZBPJ0M

src/app/_components/Gameboard/_subcomponents/Overlays/ChatDrawer/ChatDrawer.tsx
```diff
@@ -1,20 +1,83 @@
...
+
+            if (typedPlayer.cardPiles) {
+                Object.values(typedPlayer.cardPiles).forEach(pile => {
+                    if (Array.isArray(pile)) {
+                        pile.forEach(card => {
+                            if (card?.name) {
+                                const uniqueKey = `${card.name}:${playerId}`;
+                                cards[uniqueKey] = createChatCardData(card, playerId);
+                            }
+                        });
```
Frugghi, line 65: How does this work if there are multiple cards with the same name but different subtitles?

src/app/_components/_sharedcomponents/Chat/Chat.tsx
```diff
@@ -1,64 +1,358 @@
...
+                        {playerName}
+                    </Typography>:
+                    {processMessageContent(chatContent)}
+                </Typography>
+            );
+        }
+        
+        if (Array.isArray(message)) {
+            const stringMessage = message.map(item => 
+                typeof item === 'object' ? item?.name || '' : item
```
AMMayberry1, line 338: If I'm understanding right, the flow is here that we first serialize the message object into a string by reading out the card / player names, then we parse that string to figure out the structure and how to highlight.

If I'm reading it correctly, this is unnecessarily complex - all of the data we need for e.g. determining which player owns a given card is provided in the chat card data. We should figure out all of the data we need in the raw object format _before_ serializing to a string and then use that to guide how we do the highlighting, instead of reverse-engineering from the flat log string.

This has two additional benefits:

1. protects against changes to the log format, like if we refactor the attack logs to be worded differently it won't cause any issues if we use the method I'm proposing. Whereas currently such changes would break the code in the PR, which makes things difficult design-wise since the BE folks wouldn't know that some log strings are important and can't be changed

2. we don't need to iterate over all available card names and try and search each one in the string, which is an extremely expensive operation computation-wise. we would instead just get the ChatCard's ID or set code (whichever is needed) and just know which card to render

src/app/_components/_sharedcomponents/Chat/Chat.tsx
```diff
@@ -1,64 +1,358 @@
...
+    ): JSX.Element => {
+        return (
+            <Typography
+                component="span"
+                sx={{
+                    color: isCurrentPlayerCard ? 'var(--initiative-blue)' : 'var(--initiative-red)',
+                    cursor: 'pointer',
+                    fontWeight: 'bold',
+                    '&:hover': {
+                        color: 'purple',
```
AMMayberry1, line 124: This makes the highlighting the same color as the "alert" text for phase and round, which is a little visually confusing. Also, I think ideally it would still be possible to determine the controller of the card you're hovering over, so e.g. still use a shade of blue and a shade of red for the text hover color

src/app/_components/_sharedcomponents/Chat/ChatTypes.ts
```diff
@@ -1,29 +1,37 @@
...
 export interface IGameChat {
     messages: IChatEntry[];
 }
+
+export interface IChatCardData {
+    id: string;
+    name: string;
+    setId: { set: string; number: number };
+    type: string;
+    ownerId?: string;
```
AMMayberry1, line 25: For the purposes of highlighting we need to be checking the card's controller instead of its owner. This matters for units that have been stolen

PR #312, opened 2025-05-02
Add Use Force Base

danbastin:
Creates a button in NextSetPreview lobbies that sends the force base override options


AMMayberry1:
Merged the changes into the single big LOF PR

PR #328, opened 2025-05-08
#Added checks for the user object sent to the BE with each call

CheBato:
Added user object to every API call so don't need to re-get the user details on the BE.

### DIFF
--- a/src/app/_components/HomePage/_subcomponents/JoinableGame/JoinableGame.tsx
+++ b/src/app/_components/HomePage/_subcomponents/JoinableGame/JoinableGame.tsx
@@ -50,7 +50,10 @@ const JoinableGame: React.FC<IJoinableGameProps> = ({ lobby }) => {
             const payload = {
                 lobbyId: lobbyId,
                 user: { id: user?.id || localStorage.getItem('anonymousUserId'),
-                    username:user?.username || 'anonymous '+ localStorage.getItem('anonymousUserId')?.substring(0,6) },
+                    username:user?.username || 'anonymous '+ localStorage.getItem('anonymousUserId')?.substring(0,6),
+                    preferences: user?.preferences || null,
+                    welcomeMessageSeen: user?.welcomeMessageSeen || null
+                },
             };
             const response = await fetch(`${process.env.NEXT_PUBLIC_ROOT_URL}/api/join-lobby`, {
                 method: 'POST',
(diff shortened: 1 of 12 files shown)


src/app/_components/HomePage/_subcomponents/JoinableGame/JoinableGame.tsx
```diff
@@ -49,8 +49,9 @@ const JoinableGame: React.FC<IJoinableGameProps> = ({ lobby }) => {
         try {
             const payload = {
                 lobbyId: lobbyId,
-                user: { id: user?.id || localStorage.getItem('anonymousUserId'),
-                    username:user?.username || 'anonymous '+ localStorage.getItem('anonymousUserId')?.substring(0,6) },
+                user: user || { id: localStorage.getItem('anonymousUserId'),
```
AMMayberry1, line 52: Let's make a helper method for constructing the user from local storage since this code is repeated a few times

src/app/_contexts/User.context.tsx
```diff
@@ -57,9 +57,12 @@ export const UserProvider: React.FC<{ children: ReactNode }> = ({
                         provider: session.user.provider || null,
                         providerId: session.user.id || null,
                         welcomeMessageSeen: serverUser.welcomeMessageSeen,
+                        authenticated: true,
+                        preferences: serverUser.preferences
                     });
                 } catch (error) {
                     console.error('Error syncing user with server:', error);
+                    logout();
```
AMMayberry1, line 65: The logout operation is probably too heavy to live inside a catch block, maybe just set a flag here and do it inside an if statement after this block

src/app/_utils/DeckStorageUtils.ts
```diff
@@ -91,18 +95,22 @@ export const getUserFromServer = async(): Promise<{ id: string, username: string
...
 
-export const getUsernameChangeInfoFromServer = async(): Promise<{
+export const getUsernameChangeInfoFromServer = async(user: IUser): Promise<{
     canChange: boolean;
     message: string;
     typeOfMessage: string;
 }> => {
     try {
+        const payload = {
+            user: user
```
AMMayberry1, line 105: nit: since both names are the same can just do `const payload = { user };`, it's equivalent. Same note for other places in the PR that do this

AMMayberry1 (approved):

Issue #329, opened 2025-05-08
Add env var to control CORS origin for local testing


AMMayberry1:
Right now the CORS setup breaks logins in local dev mode. Need to add an environment variable that lets us remove the `domain` parameter from these calls in local dev mode: https://github.com/SWU-Karabast/forceteki-client/blob/1aa9d834eba8fa044f024ba6f059fff5b409f871/src/app/api/auth/%5B...nextauth%5D/route.ts#L26

PR #342, opened 2025-05-17
Add env var to control CORS origin for local testing

CheBato:
[- Added check for production for the domain-](https://github.com/SWU-Karabast/forceteki-client/issues/329)

Should make it easier to debug now in local.

### DIFF
--- a/src/app/api/auth/[...nextauth]/route.ts
+++ b/src/app/api/auth/[...nextauth]/route.ts
@@ -23,7 +23,7 @@ const handler = NextAuth({
         sessionToken: {
             name: 'next-auth.session-token',
             options: {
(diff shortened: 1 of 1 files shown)


src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -23,7 +23,7 @@ const handler = NextAuth({
         sessionToken: {
             name: 'next-auth.session-token',
             options: {
-                domain: '.karabast.net',
+                ...(process.env.NODE_ENV === 'production' ? { domain: '.karabast.net' } : {}),
```
Comment (AMMayberry1) [inline src/app/api/auth/[...nextauth]/route.ts:26]: In most cases it's better to do this the opposite way around and check for the value of `'development'`. That way if the variable doesn't get set in prod for some reason, we'll still get the right behavior

AMMayberry1 (approved):

PR #347, opened 2025-05-20
clear upgrade popup on click

danbastin:
- Clears upgrade popup on click when it is selectable. This might introduce a popup flicker in cases where you are not selecting the upgrade for defeat but defeating an upgrade is what we're doing majority of the time when they are selectable so I think this is a fine tradeoff. 
- Update mouse styling when hovering subcards
- Remove aria warnings in console related to hover popups


moyerr:
Just tested this on iPhone and on desktop browser, and the issue is resolved for both 🔥

AMMayberry1 (approved):

PR #354, opened 2025-05-24
Discard / Resource element scaling

brianries:
**Summary**
This PR enhances the responsiveness and visual scaling of the Discard/Draw Deck/Resources tray in the player and opponent tray components.

**Key Changes**

- Discard / Deck / Resource layout changes:
  - In portrait mode, the resources render above the discard / deck
  - discard / deck aspect ratio maintained in both narrow and wide element rendering -- this required a layout effect listener capturing the aspect ratio to properly flip the width / height auto   
  
- Resource element changes: 
  - Text size adjusts smoothly between mobile, tablet, and desktop at various breakpoints
  - On screen widths narrower than ipadPro the element will be rendered vertically with the resource icon above the numbers
  - Portrait mode retains the horizontal layout of the resource element despite screen size

- Refactor and Minor Enhancements:
  - Some more changes to the center column in various screen sizes (leader cards were rending too small with chat open etc)

All Portrait mode rendering:
image

Smaller screen size (<= 1366px) landscape mode rendering:
image

Larger width landscape mode:
image


brianries:
> Test out the resource with one player having 5 and the other having 12. I think we can move the Resource icon over the numbers a litte earlier. Once we get to double digit resources we get a bit of overflow aroun the 1700-1500px range. image

Made a new 1600px breakpoint -- called it desktopHD.  Under 1600px it will render in column style now.  Here is the element at exactly 1600, at 1599 it renders vertical

image

src/app/_components/Gameboard/Board/Board.tsx
```diff
@@ -41,7 +44,7 @@ const Board: React.FC<IBoardProps> = ({
         },  
         unitsLayer: {
             position: 'relative',
-            margin: '0 1rem',
+            margin: isPortrait ? '0 0.2em' : '0 1rem',
```
danbastin, line 47: nit: these units are preobably interchangeable for us, but lets just stay consistent at least at the property level. Either one is fine here.

danbastin (approved):

PR #368, opened 2025-06-01
Che bato/leader flip side hover

CheBato:
- Possible implementation for flipside swap with ctrl.

### DIFF
--- a/src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
+++ b/src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
@@ -45,6 +45,8 @@ const LeaderBaseCard: React.FC<ILeaderBaseCardProps> = ({
     const handlePreviewClose = () => {
         clearTimeout(hoverTimeout.current);
         setAnchorElement(null);
+        setLeaderBackgroundImage(null);
+        setIsCtrl(false);
     };
 
     const defaultClickFunction = () => {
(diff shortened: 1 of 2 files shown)


src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
```diff
@@ -1,28 +1,69 @@
...
     const { sendGameMessage, connectedPlayer, getConnectedPlayerPrompt, distributionPromptData, gameState } = useGame();
-    
+    const [leaderBackgroundImage, setLeaderBackgroundImage] = React.useState<string | null>(null);
+    const [isCtrl, setIsCtrl] = React.useState<boolean>(false);
     const [anchorElement, setAnchorElement] = React.useState<HTMLElement | null>(null);
     const hoverTimeout = React.useRef<number | undefined>(undefined);
     const open = Boolean(anchorElement);
-    
+
+    useEffect(() => {
```
danbastin, line 23: Since we are using this in 3 places we might see if we can pull these out into a custom hook. This is just a function wrapper for the useEffect where the function name starts with "use". Export that from a seperate file (we currently do not have a hooks folder so this would be a good time to create one at src/app/_hooks) and call it in each spot with the necessary parameters.

src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
```diff
@@ -19,50 +20,15 @@ const LeaderBaseCard: React.FC<ILeaderBaseCardProps> = ({
...
-            };
-            window.addEventListener('keydown', handleKeyDown);
-            window.addEventListener('keyup', handleKeyUp);
-            return () => {
-                window.removeEventListener('keydown', handleKeyDown);
-                window.removeEventListener('keyup', handleKeyUp);
-            };
-        }
-    }, [anchorElement, card, cardStyle, isLeader]);
+    console.log(card?.setId);
```
danbastin, line 23: Much better organization. I really like this set up. Lets clean out the conosle.logs and I think this is good to go.

danbastin (approved):

PR #369, opened 2025-06-01
Add Warning/Danger Timer on Board

Trys2Hard:


src/app/_components/Gameboard/OpponentCardTray/OpponentCardTray.tsx
```diff
@@ -65,6 +68,36 @@ const OpponentCardTray: React.FC<IOpponentCardTrayProps> = ({ trayPlayer, prefer
         };
     }
 
+    const pulseYellowTimer = keyframes`
+  0% {
```
danbastin, line 72: indentation here

src/app/_components/Gameboard/OpponentCardTray/OpponentCardTray.tsx
```diff
@@ -178,6 +219,7 @@ const OpponentCardTray: React.FC<IOpponentCardTrayProps> = ({ trayPlayer, prefer
                     ...styles.rightColumn,
                 }}
             >
+                <AccessAlarm sx={styles.timer}/>
```
danbastin, line 222: I think we can make this look a little cleaner if we wrap this icon in a `Box` and apply an backgroung gradient to that box. We should be able to make that gradient go from transparent -> color -> transparent so that the glow looks more like a ring bihind the rim of the clock instead of being outside of the icon.

src/app/_components/Gameboard/OpponentCardTray/OpponentCardTray.tsx
```diff
@@ -133,6 +162,15 @@ const OpponentCardTray: React.FC<IOpponentCardTrayProps> = ({ trayPlayer, prefer
             backgroundRepeat: 'no-repeat',
             aspectRatio: '1 / 1.4',
             width: '16rem',
+        },
+        timerBox: {
+            display: !warning && !danger ? 'none' : 'block',
+            borderRadius: '50%',
+            animation: warning ? `${pulseYellowTimer} 3s infinite ease-in-out` : danger ? `${pulseRedTimer} 3s infinite ease-in-out` : 'transparent',
+        },
+        timer: {
```
danbastin, line 171: lets add `display:block` to this. The inline defualt styling is causing a bit of lineheight weirdness making the box a little taller than square which is what is warping the drop shadow just a touch.

danbastin (approved):

Issue #295, opened 2025-04-13
Add MjRToland's clicking sound for in-game actions


AMMayberry1:
MjrToland created a clicking sound effect for us we can us for doing game actions:

https://drive.google.com/file/d/15ULgfNOL7S5vfUNZbN7EgCXcxQ6L8S5y/view?usp=drive_link

We need to add this to our basic stuff like button clicks, chat messages (from a player), etc. Very heavily-requested feature.

CheBato:
need BE support for resourcing.

PR #371, opened 2025-06-04
#295 Add click sound to card clicks, chat messages and hello there sound when a user joins

CheBato:
- Added click and hello there sounds to the platform
- Supports Anon preferences and Logged in users preferences.
- Added warning if you change preferences and want to leave.
- Added sync when a user logs in

BE PR: https://github.com/SWU-Karabast/forceteki/pull/1488


AMMayberry1:
I think the "Save Changes" button is a little too subtle when it's flipped into the "on" state, might be better if we use the green color we've used for other buttons to make it a little more obvious

AMMayberry1:
Sound preferences should be available from the in-game menu as well

AMMayberry1:
Mentioned in chat but leaving a note here - I think the idea was to also make sound preferences available to be configured by anon users. Ideally we can set up some data structures that make it seamless to interact with user settings regardless of whether they're local storage or dynamodb

AMMayberry1:
Also mentioned in chat, but imo the sound for when you are being prompted for your action is the single most important one (it's the one we get asked about the most). Would be good to get that added, unless there's some issue with it

AMMayberry1:
I think the player message sound should be something different than the clicking sound so that it's easier to catch

AMMayberry1:
For preferences, I think "Clicking on cards" and "Clicking on menu buttons" don't need to be two separate settings. It can be one for "clicking on prompts and cards", those are such similar actions from a UI perspective that I don't think anyone would care to enable one but not the other

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
```diff
@@ -0,0 +1,270 @@
...
+        },
+        saveButton: {
+            minWidth: '120px',
+        },
+    };
+
+    const handleVolumeChangeCommitted = (value: number) => {
+        const volumeLevel = value / 100;
+        try {
+            const tempAudio = new Audio('/HelloThere.mp3');
```
AMMayberry1, line 62: This object is probably a little expensive to construct, so let's maybe build it outside of this hook so it's only created on page load

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
```diff
@@ -0,0 +1,270 @@
...
+    const [soundPreferences, setSoundPreferences] = useState<Preferences['sound']>({
+        muteAllSound: false,
+        volume: 0.75,
+        muteCardClickSound: false,
+        muteMenuButtonsSound: false,
+        muteChatSound: false,
+        muteOpponentFoundSound: false,
+    });
+    const [isSaving, setIsSaving] = useState(false);
+    const [saveStatus, setSaveStatus] = useState<'idle' | 'success' | 'error'>('idle');
```
AMMayberry1, line 31: Can we use a string type (e.g. `type SaveStatus = 'idle' | 'success' | 'error'`) or an enum here?

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
```diff
@@ -0,0 +1,270 @@
...
+        if (!originalPreferences) return;
+
+        const hasUnsavedChanges = JSON.stringify(soundPreferences) !== JSON.stringify(originalPreferences);
+        setHasChanges(hasUnsavedChanges);
+        if (setHasNewChanges) {
+            setHasNewChanges(hasUnsavedChanges)
+        }
+    }, [soundPreferences, originalPreferences]);
+
+    const handlePreferenceChange = (key: string, value: boolean | number) => {
```
AMMayberry1, line 105: With some type magic we can restrict the parameter types here to guarantee that they are always valid. First we need to refactor the `Preferences` type a little bit. This step might be optional but I couldn't find a way around it.

```ts
export interface ISoundPreferences {
    muteAllSound?: boolean;
    volume?: number;
    muteCardClickSound?: boolean;
    muteMenuButtonsSound?: boolean;
    muteChatSound?: boolean;
    muteOpponentFoundSound?: boolean;
};

export interface Preferences {
    cardback?: string;
    sound?: ISoundPreferences;
}
```

Then we can change the function signature here:

```ts
const handlePreferenceChange = <T extends keyof ISoundPreferences>(key: T, value: ISoundPreferences[T]) => {
```

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
```diff
@@ -0,0 +1,270 @@
...
+        setSoundPreferences(prev => ({
+            ...prev,
+            [key]: key === 'volume' ? value as number / 100 : value
+        }));
+        setSaveStatus('idle');
+    };
+
+    const handleSavePreferences = async (): Promise<boolean> => {
+        setIsSaving(true);
+        setSaveStatus('idle');
```
AMMayberry1, line 115: What does 'idle' mean in this context? No change?

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+        try {
+            if (user) {
+                const success = await saveSoundPreferencesToServer(user, soundPreferences);
```
AMMayberry1, line 119: Since we will be having a lot of preferences that will be mirrored between local storage and the server, could we come up with some generic helper methods for handling a preference update regardless of which user type it is?

It would be cool if we could get as far as having a method that doesn't need to be updated every time we add new settings by using maybe `Partial<Preferences>` or something like that so you just provide an object that has the specific preferences to be updated. But there might be cases where we require a specific set of preferences that have to be provided together so probably we can't have something quite that generic sadly (though I leave the decision to you if you think we can).

src/app/_components/_sharedcomponents/Preferences/_subComponents/VerticalTabs.tsx
```diff
@@ -27,11 +29,46 @@ function VerticalTabs({
...
 
+    useEffect(() => {
+        const handleBeforeUnload = (e: BeforeUnloadEvent) => {
+            if (tabs[value] === 'soundOptions' && hasUnsavedChanges) {
+                e.preventDefault();
+                e.returnValue = 'You have unsaved sound preferences. Are you sure you want to leave?';
+            }
+        };
+
+        window.addEventListener('beforeunload', handleBeforeUnload);
```
AMMayberry1, line 46: nit: if we can use camel case for the event name that'd be more consistent with our other events

src/app/_components/_sharedcomponents/Preferences/_subComponents/VerticalTabs.tsx
```diff
@@ -42,8 +79,8 @@ function VerticalTabs({
                 return <KeyboardShortcutsTab/>;
             case 'cardSleeves':
                 return <CardSleevesTab/>;
-            case 'gameOptions':
-                return <GameOptionsTab/>;
+            case 'soundOptions':
```
AMMayberry1, line 82: It might be slightly outside the scope of this PR but it would be really good to get all of these strings defined in a type or enum as well, for both this and renderLabels below

src/app/_hooks/useSoundHandler.tsx
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }, [volume]);
+
+    // Main function to play sounds
+    const playSound = useCallback((action: SoundAction) => {
+        if (!enabled || preferences?.muteAllSound) {
+            return;
+        }
+        // Check specific action mutes
+        const actionMutes = {
```
AMMayberry1, line 75: Do we need to define this object here? Could just do `if (preferences?.[action])` below

AMMayberry1:
Summarizing from discord discussions about sounds:
- Use rotating beep noises for player chat messages
- Toland will gave us a normal-speed "Hello There" sound to try
- Need a different sound for the "it's your action" noise, not sure what yet

AMMayberry1:
The property to check for when to play the "it's your action" sound is `Player.promptState.playerIsNewlyActive`

src/app/_components/Lobby/Players/Players.tsx
```diff
@@ -1,17 +1,32 @@
...
     // ------------------------STYLES------------------------//
     const { connectedPlayer, lobbyState } = useGame();
+    const { user } = useUser();
     const connectedUser = lobbyState ? lobbyState.users.find((u: ILobbyUserProps) => u.id === connectedPlayer) : null;
     const opponentUser = lobbyState ? lobbyState.users.find((u: ILobbyUserProps) => u.id !== connectedPlayer) : null;
+    const previousOpponentRef = useRef<ILobbyUserProps | null>(null);
+    const { playFoundOpponentSound } = useSoundHandler({ user });
+    useEffect(() => {
+        // Check if opponent just joined (was null/undefined, now exists)
+        const hadOpponent = opponentUser?.id === previousOpponentRef.current;
```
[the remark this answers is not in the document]
CheBato: This is actually if the opponent refreshes. They "don't show as disconnected", but the sound plays since the value changed.

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
CheBato: OH! So cool great idea

src/app/_components/QuickGame/FoundGame/FoundGame.tsx
```diff
@@ -32,6 +41,14 @@ const FoundGame: React.FC = () => {
         }
     }, [router, gameState, lobbyState]);
 
+    useEffect(() => {
+        if (!hasPlayedSoundRef.current) {
+            playFoundOpponentSound();
+            hasPlayedSoundRef.current = true;
```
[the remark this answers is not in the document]
AMMayberry1: Oh that's my mistake, I was confusing this with a different ref

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
AMMayberry1: The meaning is a little unclear, maybe something more explicit like `noChange` would be good

src/app/_hooks/useSoundHandler.tsx
AMMayberry1: Unless I'm mistaken it looks like this is only referenced on line 84, so I'm saying we could just consolidate to look like this:

```ts
if (preferences?.[action]) {
    return;
}
```

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
```diff
@@ -11,27 +11,55 @@ import {
...
+import { keyframes } from '@mui/system';
+import { savePreferencesGeneric } from '@/app/_utils/genericPreferenceFunctions';
+import { useSoundHandler } from '@/app/_hooks/useSoundHandler';
 
 interface SoundOptionsTabProps {
     setHasNewChanges?: Dispatch<SetStateAction<boolean>>;
 }
+
+enum SaveStatus {
+    NOCHANGE = 'noChange',
```
AMMayberry1, line 25: nit: we'll typically name the enums with camel case, so `NoChange = 'noChange'`

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
```diff
@@ -11,27 +11,55 @@ import {
...
     });
+
+    const [soundPreferences, setSoundPreferences] = useState<IPreferences['sound']>(() => {
+        return getPreferences().sound
+    });
+
     const [isSaving, setIsSaving] = useState(false);
-    const [saveStatus, setSaveStatus] = useState<'idle' | 'success' | 'error'>('idle');
+    const [saveStatus, setSaveStatus] = useState<SaveStatus>(SaveStatus.NOCHANGE);
+    const tempAudio = new Audio('/HelloThere.mp3');
```
AMMayberry1, line 58: nit: let's move this line out below the react state declarations since it's a bit different

src/app/_components/_sharedcomponents/Preferences/_subComponents/VerticalTabs.tsx
```diff
@@ -25,6 +25,17 @@ function tabProps(index: number) {
 }
 
 
+enum TabType {
+    CURRENT_GAME = 'currentGame',
```
AMMayberry1, line 29: nit: same note here, our convention would typically be `CurrentGame = 'currentGame'`

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
```diff
@@ -0,0 +1,287 @@
...
+                        onChange={(muted) => handlePreferenceChange('muteCardAndButtonClickSound', muted)}
+                        defaultChecked={soundPreferences?.muteCardAndButtonClickSound}
+                        disabled={soundPreferences?.muteAllSound}
+                    />
+                </Box>
+
+                <Box sx={styles.optionContainer}>
+                    <PreferenceOption
+                        option={'Your turn indicator'}
+                        optionDescription={'Sound when it is your turn to perform an action.'}
```
AMMayberry1, line 232: nit: I would tweak this slightly to be "Sound when you need to perform an action", since it's not just on your turn

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/SoundOptionsTab.tsx
```diff
@@ -0,0 +1,287 @@
...
+            </Box>
+
+            <Box sx={styles.functionContainer}>
+                <Typography sx={styles.typographyContainer} variant={'h2'}>In-game Sound</Typography>
+                <Divider sx={{ mb: '20px' }}/>
+
+                <Box sx={styles.optionContainer}>
+                    <PreferenceOption
+                        option={'Card & Button Clicks'}
+                        optionDescription={'Sound when clicking on cards, buttons and prompts.'}
```
AMMayberry1, line 221: really minor nit: since these descriptions aren't complete sentences, they shouldn't end in a period

danbastin (approved):

PR #374, opened 2025-06-06
#Palpatine preview fix

CheBato:


src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
```diff
@@ -20,13 +20,21 @@ const LeaderBaseCard: React.FC<ILeaderBaseCardProps> = ({
     const [anchorElement, setAnchorElement] = React.useState<HTMLElement | null>(null);
     const hoverTimeout = React.useRef<number | undefined>(undefined);
     const open = Boolean(anchorElement);
+    let startingSide = card?.onStartingSide
```
danbastin, line 23: Something we can do is move all of this logic into the hook and just export out what is needed to render (like the aspect ratio). That way we can keep the cards themselves a little cleaner and whatever solution we come up with to handle all the cases can just be contained in the hook.

danbastin (approved):

Issue #360, opened 2025-05-29
New player window populates username with a real value


AMMayberry1:
In the new player window it populates it with a real string value, so if you click "set username" you get an error saying that it's the current username:

![Image](https://github.com/user-attachments/assets/a6402a34-4360-434c-83d8-ac34868612e5)

This is a bit confusing. Maybe it should do something like "Your current username is [X], you can change it now:" and then that username appears in the box as a suggestion instead of an actual value. Lmk if it's unclear what I mean

PR #376, opened 2025-06-10
# Improved Token security and management.

CheBato:
- We now add the uuid of the user to the session and JWT token once we authorize them.
- We recreate and sign the JWT token when we update with the uuid.

BE PR:
https://github.com/SWU-Karabast/forceteki/pull/1406

### DIFF
--- a/src/app/_components/_sharedcomponents/HomescreenWelcome/WelcomePopup.tsx
+++ b/src/app/_components/_sharedcomponents/HomescreenWelcome/WelcomePopup.tsx
@@ -24,7 +24,6 @@ const WelcomePopup: React.FC<WelcomePopupProps> = ({ open, onClose }) => {
     useEffect(() => {
         if (open) {
             const initialUsername = user?.username || '';
-            // setUsername(initialUsername); // Pre-fill the username input
             const validationError = validateDiscordUsername(initialUsername);
             setUserErrorSummary(validationError);
             setCanSubmitUsername(validationError === null && initialUsername.trim() !== '');
--- a/src/app/_components/_sharedcomponents/HomescreenWelcome/WelcomePopup.tsx
+++ b/src/app/_components/_sharedcomponents/HomescreenWelcome/WelcomePopup.tsx
@@ -24,7 +24,7 @@ const WelcomePopup: React.FC<WelcomePopupProps> = ({ open, onClose }) => {
     useEffect(() => {
         if (open) {
             const initialUsername = user?.username || '';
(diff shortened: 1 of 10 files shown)


src/app/_components/_sharedcomponents/HomescreenWelcome/WelcomePopup.tsx
```diff
@@ -24,7 +24,7 @@ const WelcomePopup: React.FC<WelcomePopupProps> = ({ open, onClose }) => {
     useEffect(() => {
         if (open) {
             const initialUsername = user?.username || '';
-            setUsername(initialUsername); // Pre-fill the username input
+            // setUsername(initialUsername); // Pre-fill the username input
```
AMMayberry1, line 27: Was this commented out for testing?

src/app/_utils/DeckStorageUtils.ts
```diff
@@ -59,7 +59,7 @@ export const retrieveDecksForUser = async <T extends 'stored' | 'display' = 'sto
         }
     } catch (err) {
         console.error('Error fetching decks:', err);
-        throw err;
+        loadSavedDecks();
```
AMMayberry1, line 62: What's the reason for this change?

src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -83,6 +86,7 @@ const handler = NextAuth({
                 email: token.email || session.user?.email || null,
                 image: token.picture || session.user?.image || null,
                 provider: token.provider as string || 'unknown',
+                uuid: token.uuid ?? null
```
Comment (AMMayberry1) [inline src/app/api/auth/[...nextauth]/route.ts:89]: For both PRs - could we name this something a little more specific like "userId"

src/app/_components/_sharedcomponents/HomescreenWelcome/WelcomePopup.tsx
CheBato: no it was for the new Player window populates username issue you wrote

src/app/_utils/DeckStorageUtils.ts
AMMayberry1: Ah I see, sounds like this is the cause for that comment we're discussing on the BE side. I think we should investigate this further, it's sort of quietly covering up the problem without really addressing it. And it's also possible that there are other reasons we could get an exception which would now be quietly hidden.

I'll follow up in chat to discuss potential solutions

AMMayberry1:
Resolves #360

danbastin (approved):

PR #381, opened 2025-06-14
Check oppoennt zones for multiselect

danbastin:
Cards allowing selection in opponents other zones were not properly checking those zones. This change makes sure those zones are being checked and that selection of cards in those zones appropriately updates the popup.

Fixes issues with Restock prompt UI when selecting from opponent's discard.


AMMayberry1:
It feels inconsistent that the selection border is red when selecting from opponent's discard

image

AMMayberry1:
Looks like discard window doesn't refresh or close when a selection is made, so it looks like nothing happened (and it's hard to select the target on the board)

AMMayberry1 (approved):

PR #417, opened 2025-07-08
Attempt new socket connection if queue is timing out

AMMayberry1:
Another stab at fixing the connection issues, this time from the FE side. If the FE detects that it is not receiving the queue heartbeat, it will aggressively try reconnecting to the BE in hopes that it can catch the lobby before it times out.


src/app/_contexts/Game.context.tsx
```diff
@@ -79,126 +200,6 @@ export const GameProvider = ({ children }: { children: ReactNode }) => {
             auth: token ? { token } : undefined,
         });
 
-        const cardSelectableZones = (gamestate: any) => {
```
danbastin, line 82: Thanks for cleaning this effect up.

PR #435, opened 2025-08-05
# New feature prompt generation

CheBato:
- Added a new prompt that is set based on localstorage.


src/app/_components/_sharedcomponents/HomescreenWelcome/NewFeaturePopup.tsx
```diff
@@ -0,0 +1,114 @@
...
+            aria-labelledby="whats-new-dialog-title"
+            sx={styles.dialog}
+        >
+            <DialogContent>
+                <Typography variant="h4" sx={styles.title} id="whats-new-dialog-title">
+                    ✨ New Feature - SWUStats
+                </Typography>
+
+                <Box component="ul" sx={styles.whatsNewList}>
+                    <li>
```
AMMayberry1, line 89: Could we make this not be a list? The single bullet looks a bit strange imo:

image

src/app/_components/_sharedcomponents/HomescreenWelcome/NewFeaturePopup.tsx
```diff
@@ -0,0 +1,114 @@
...
+                        <Typography component="span" variant="body2" sx={styles.whatsNewItem}>
+                            Track your wins, losses, and match‑up win‑rates on <strong>SWUSTATS</strong>! By adding a deck on swustats and importing it into Karabast
+                            you are able to retrieve in-depth information on your decks performance!
```
AMMayberry1, line 92: nit, typo: should be "deck's performance" instead of "decks performance". Also we probably won't roll this announcement out until we have card metrics so might be good to mention something about that

PR #442, opened 2025-08-17
chebato/unification stats

CheBato:
FE PR for stats consolidation.
BE PR: https://github.com/SWU-Karabast/forceteki/pull/1580


src/app/DeckPage/[DeckId]/page.tsx
```diff
@@ -436,6 +477,15 @@ const DeckDetails: React.FC = () => {
                         <Typography variant="h3" sx={styles.titleText}>
                             {deckData?.metadata.name}
                         </Typography>
+                        { opponentStats && (
+                            <Box>
```
Comment (AMMayberry1) [inline src/app/DeckPage/[DeckId]/page.tsx:481]: The button needs to be much more prominent, I was looking for it and it took me a minute to find it. Most people wouldn't see it at all.

Imo probably best to just make it a full button with text like "Download as CSV" or something, but there could be other options if you have an idea you like

src/app/DeckPage/[DeckId]/page.tsx
```diff
@@ -228,6 +229,43 @@ const DeckDetails: React.FC = () => {
...
+            'Losses',
+            'WinPercentage',
+        ];
+
+        const rows = opponentStats.map(s => [
+            s.leaderId,
+            s.baseId,
+            s.wins,
+            s.losses,
+            `${s.winPercentage}%`,
```
Comment (AMMayberry1) [inline src/app/DeckPage/[DeckId]/page.tsx:249]: I think we probably shouldn't include the % character in here, that just makes it harder to parse

src/app/DeckPage/[DeckId]/page.tsx
```diff
@@ -228,6 +229,43 @@ const DeckDetails: React.FC = () => {
...
+        ]);
+
+        // build CSV
+        const escape = (v: unknown) => {
+            const s = String(v ?? '');
+            return /[",\n]/.test(s) ? `"${s.replace(/"/g, '""')}"` : s;
```
Comment (AMMayberry1) [inline src/app/DeckPage/[DeckId]/page.tsx:255]: What is this doing? Having a bit of trouble following it

Comment (CheBato) [reply src/app/DeckPage/[DeckId]/page.tsx:255]: oh so what this does is checks if the format is correct of the values in case we ever get situations such as ""name"" -> and this needs to be then "\"name\"" for instance. Thought it would be so we have it made clean :)

src/app/DeckPage/[DeckId]/page.tsx
```diff
@@ -228,6 +229,43 @@ const DeckDetails: React.FC = () => {
 
     const isSmallScreen = useMediaQuery('(max-width: 1280px)');
 
+    // add near your other handlers
+    const exportOpponentStatsCSV = () => {
+        if (!opponentStats || opponentStats.length === 0) return;
+
+        const header = [
+            'OpponentLeaderId',
```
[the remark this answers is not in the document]
Comment (CheBato) [reply src/app/DeckPage/[DeckId]/page.tsx:237]: I actually gave them an example of the format:
https://discord.com/channels/1220057752961814568/1280363500178505829/1406564027525496833

melee would be best probably but i'd have to retrieve the ids and the melee names. Is there a function for this in the BE?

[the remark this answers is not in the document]
Comment (AMMayberry1) [reply src/app/DeckPage/[DeckId]/page.tsx:237]: Not yet, we should add one and then those can be sent along with the deck data so we don't have to make any followup requests

src/app/_components/DeckPage/DeckComponent/AnimatedStatsTable.tsx
```diff
@@ -243,11 +244,13 @@ const AnimatedStatsTable: React.FC<AnimatedStatsTableProps> = ({
                             <TableCell sx={{ borderBottom: 'none' }}>
                                 <Box sx={styles.leaderBaseHolder}>
                                     <Box sx={styles.CardSetContainerStyle}>
-                                        {row.baseId?.startsWith('Base') ? (
+                                        {row.baseId?.startsWith('30') || row.baseId?.startsWith('28') ? (
```
AMMayberry1, line 247: nit: maybe match on a little bit more of the string in case there's ever a base that has name that actually starts with one of these numbers

src/app/DeckPage/[DeckId]/page.tsx
```diff
@@ -436,6 +486,15 @@ const DeckDetails: React.FC = () => {
                         <Typography variant="h3" sx={styles.titleText}>
                             {deckData?.metadata.name}
                         </Typography>
+                        { opponentStats && (
+                            <Box>
+                                <Tooltip title="Download as CSV">
+                                    <Button sx={styles.iconButton} onClick={exportOpponentStatsCSV}>
+                                        Download as csv
```
Comment (AMMayberry1) [inline src/app/DeckPage/[DeckId]/page.tsx:493]: To keep the casing consistent within the page, should be "Download as CSV"

src/app/_components/DeckPage/DeckComponent/AnimatedStatsTable.tsx
```diff
@@ -227,23 +241,42 @@ const AnimatedStatsTable: React.FC<AnimatedStatsTableProps> = ({
...
                         <TableRow key={index}>
-                            <TableCell sx={{ borderBottom:'none' }}>
+                            <TableCell sx={{ borderBottom: 'none' }}>
                                 <Box sx={styles.leaderBaseHolder}>
                                     <Box sx={styles.CardSetContainerStyle}>
-                                        <Box>
-                                            <Box sx={{ ...styles.boxGeneralStyling, backgroundImage:`url(${s3CardImageURL({ id: row.baseId, count:0 })})` }}
+                                        {row.baseId?.startsWith('30hp') || row.baseId?.startsWith('28hp') ? (
+                                            <Box sx={{ ...styles.parentBoxStyling, left: '70px', top: '25px', zIndex:'1' }}>
+                                                <Tooltip title={row.baseId?.startsWith('30') ? 'A basic 30hp base' : 'A basic 28hp force base'}>
```
AMMayberry1, line 249: nit: idk if it's actually grammatically better but I think it reads better as "Basic 28hp force base" (no "a" at the beginning), same for the non-force base

PR #447, opened 2025-09-02
Add a sort order + filterByLeader to public games to spectate

phever:
My first PR! 🎉

Added a sort by oldest/newest dropdown to the public games. As well as a filter to see only games of a certain leader. Also removed a line of filtering that has been moved to the backend.


phever:
image

New preview without the ascending/descending option.

AMMayberry1:
I'm not sure how much control we have over the styling but there's a few things that I think could be rendered a little more clearly. The miniaturized "Filter by Leader" text here is hard to spot when it's blue like this, and even when white it's so small that it's probably unreadable on phones. I think it would be good to have some separate floating text that explains what the box does, or something like that:

image

Additionally, the "X" button to clear out the value doesn't show up unless you hover it, which is problematic on phones

phever:
> I'm not sure how much control we have over the styling but there's a few things that I think could be rendered a little more clearly. The miniaturized "Filter by Leader" text here is hard to spot when it's blue like this, and even when white it's so small that it's probably unreadable on phones. I think it would be good to have some separate floating text that explains what the box does, or something like that:
> image
> 
> Additionally, the "X" button to clear out the value doesn't show up unless you hover it, which is problematic on phones. Again might be better to have an explicit "clear" button or something like that
> image

Happy to add/remove/style whatever UI elements you'd like. I'm very new to the project though so I'm not sure what to do in regards "mights" or "something like thats." I will leave it as is until I know what is required.

phever:
Got a bunch of feedback regarding direction. Thank you!!

Should be ready for a review now.

Issue #457, opened 2025-09-09
Add game count to leader filter entries


AMMayberry1:
For each leader in the filter on the main page, it would be good to show the number of games for that leader on the right-hand side so you can tell easily which leaders actually have games to see

PR #459, opened 2025-09-09
add leader filter count

phever:
Add a ($ongoingGames) to the end of each leaders filter.

image

for #457

### DIFF
--- a/src/app/_components/HomePage/_subcomponents/GamesInProgress/GamesInProgress.tsx
+++ b/src/app/_components/HomePage/_subcomponents/GamesInProgress/GamesInProgress.tsx
@@ -202,6 +202,16 @@ const GamesInProgress: React.FC = () => {
                     }
                 }
             } 
+        },
+        leaderName: {
+            flexGrow: 1,
+            '& span': {
+                color: '#bbbbbb',
+            },
+        },
+        leaderActiveGamesCount: {
+            color: '#fff',
+            fontWeight: 600,
         }
     };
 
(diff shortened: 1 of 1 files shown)


src/app/_components/HomePage/_subcomponents/GamesInProgress/GamesInProgress.tsx
```diff
@@ -55,6 +56,27 @@ const fetchLeaderData = async (setLeaderData: (leaders: LeaderNameData[] | null)
     }
 }
 
+const updateLeaderData = (
+    games: OngoingGamesData | null,
+    leaderData: LeaderNameData[] | null,
+    setLeaderData: (leaders: LeaderNameData[] | null) => void
+) => {
+    if (!leaderData) return;
+    const newLeaderData = leaderData.map(leader => ({
```
danbastin, line 65: Not super important I think since we're not dealing with large data sets here, but it may be better to just count up leader occourances in a Set and built the leaderData from that so you only have to iterate games once instead of for each leader.

src/app/_components/HomePage/_subcomponents/GamesInProgress/GamesInProgress.tsx
```diff
@@ -55,6 +56,27 @@ const fetchLeaderData = async (setLeaderData: (leaders: LeaderNameData[] | null)
...
+            game => game.player1Leader.id === leader.id || game.player2Leader.id === leader.id
+        ).length ?? 0
+    }));
+    setLeaderData(newLeaderData);
+}
+
+const getOptionLabel = (option: LeaderNameData): string => {
+    const count = option.ongoingGamesCount ?? 0;
+    const subtitle = option.subtitle ? ` - ${option.subtitle}` : '';
+    return `${option.name}${subtitle} (${count})`;
```
AMMayberry1, line 77: Would it be possible to right-justify the count part? I think it would look a little cleaner and easier to read that way, but not a big deal if it's a pain. But something to visually distinguish the number from the leader name would be great

phever: 250909_16h36m41s_screenshot

How does this look?

AMMayberry1: Fantastic!!

AMMayberry1 (approved):

PR #516, opened 2025-10-16
chebato/swustats-accesstoken-fix

CheBato:
- Changed the refershToken to boolean field as the main indicator of whether we have swustats linked or not.

BE PR: https://github.com/SWU-Karabast/forceteki/pull/1842


src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GeneralTab.tsx
```diff
@@ -87,6 +91,27 @@ function GeneralTab() {
...
+        if (user) {
+            setIsSWUStatsInCheck(true);
+            try {
+                const linked = await checkSwuStatsLinkStatus(user);
+                setIsSWUStatsLinked(linked);
+            } catch (error) {
+                console.error('Failed to check SWUStats link status:', error);
+                setIsSWUStatsLinked(false);
+            } finally {
+                setTimeout(() => {
```
AMMayberry1: What's the reason for this? If the request legitimately takes longer than a second, it's probably better to just let it spin rather or show an error message after some long period of time. We shouldn't ever show a value that we can't guarantee is correct

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GeneralTab.tsx
```diff
@@ -120,6 +144,9 @@ function GeneralTab() {
         const currentUsername = user?.username || '';
         setUsername(currentUsername);
         getUsernameChangeInfo();
+        checkSwuStatsLink().finally(
```
AMMayberry1: Why the empty `finally`?

CheBato: i wanted to add the timeout here but then i rather did in the function

src/app/_utils/ServerAndLocalStorageUtils.ts
```diff
@@ -543,6 +543,42 @@ export const getDeckFromServer = async (deckId: string, user:IUser): Promise<IDe
...
+                credentials: 'include'
+            }
+        );
+
+        if (!response.ok) {
+            // Handle authentication errors gracefully
+            if (response.status === 401) {
+                return false;
+            }
+            throw new Error('Failed to check SWUStats link status');
```
AMMayberry1: Should we instead show an error message where the button goes? Throwing an error like this will break the page, right?

I'm asking b/c we are concretely having errors at this step due to swustats going down or token issues - if those would result in us getting here then we should make sure we can handle them in a way that's intuitive for the user

PR #520, opened 2025-10-18
Final changes for undo release

AMMayberry1:
Few odds and ends for the public undo release! Major changes:

- Added a popup explaining the rules for undo in public games (see below) and state tracking to avoid repeating it
- Switched the meaning of the undo setting in private lobbies to be an option to switch from totally free undo (default) to a "request undo" mode like public games
- New styles for the undo button depending on state: none available, request undo available, free undo available, or blocked by opponent

undopopup

### DIFF
--- a/src/app/_components/_sharedcomponents/HomePagePlayMode/UndoTutorialPopup.tsx
+++ b/src/app/_components/_sharedcomponents/HomePagePlayMode/UndoTutorialPopup.tsx
@@ -56,6 +56,16 @@ const UndoTutorialPopup: React.FC<IUndoTutorialPopupProps> = ({ open, onClose })
             mt: '12px',
             display: 'flex',
             justifyContent: 'center',
+            flexWrap: 'wrap',
+            gap: '16px',
+            alignItems: 'center',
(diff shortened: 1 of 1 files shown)


src/app/_components/Gameboard/_subcomponents/Overlays/ChatDrawer/ChatDrawer.tsx
```diff
@@ -53,45 +87,83 @@ const ChatDrawer: React.FC<IChatDrawerProps> = ({ sidebarOpen, toggleSidebar })
...
+
+    let buttonText;
+    switch (quickUndoState) {
+        case QuickUndoAvailableState.RequestUndoAvailable:
+            buttonText = 'Request';
+            break;
+        case QuickUndoAvailableState.UndoRequestsBlocked:
+            buttonText = 'Blocked';
+            break;
+        default:
```
brianries: Are we good with showing a normal undo button when there is no snapshot available?  Is it just a no-op?

brianries (approved):

PR #528, opened 2025-10-23
#JSON-error-message

CheBato:
- Added more clear messaging for JSON imports.

### DIFF
--- a/src/app/_components/Lobby/_subcomponents/SetUpCard/SetUpCard.tsx
+++ b/src/app/_components/Lobby/_subcomponents/SetUpCard/SetUpCard.tsx
@@ -5,7 +5,8 @@ import {
     Typography,
     TextField,
     Button,
-    Box, CardActions, Link, Tooltip, MenuItem, Checkbox, FormControlLabel, Divider
+    Box, CardActions, Link, Tooltip, MenuItem, Checkbox, FormControlLabel, Divider,
+    FormControl
 } from '@mui/material';
 import { Info } from '@mui/icons-material';
 import { useGame } from '@/app/_contexts/Game.context';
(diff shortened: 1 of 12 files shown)


src/app/_components/_sharedcomponents/CreateGameForm/CreateGameForm.tsx
```diff
@@ -392,7 +394,7 @@ const CreateGameForm: React.FC<ICreateGameFormProps> = ({
                                 </Tooltip>
                                 )
                                 <br />
-                                OR paste deck JSON directly
+                                OR paste deck JSON directly (we do <strong>not</strong> support saving JSON)
```
AMMayberry1: Now that I'm looking at it, this is making that text box pretty verbose. Also, we do need to have an automated check here of some sort to prevent them from submitting, most people won't notice this text and then be confused why their deck didn't save.

So maybe let's remove this text from the QuickGame / CreateGame forms and just let the automated checks do the talking. I would say we should do one of the following:

1. on text entry to the field, we could do a quick check to see if it's json-formatted and if so we disable the check box and maybe adjust the text to be some message saying that it's not supported

2. we just add the same error popup as we have on the deck page if they try to save the deck

#1 is a bit more work but is probably a slightly better experience for the user since it gives immediate feedback and they're saved a click. up to you, I think both work fine

src/app/_components/Lobby/_subcomponents/SetUpCard/SetUpCard.tsx
```diff
@@ -81,6 +82,15 @@ const SetUpCard: React.FC<ISetUpProps> = ({
         }
     };
 
+    const handleJsonDeck = (deckLink: string) => {
+        const parsedInput = parseInputAsDeckData(deckLink);
+        if(parsedInput.type === 'json'){
+            setIsJsonDeck(true)
```
AMMayberry1: nit: missing semicolon

src/app/_components/_sharedcomponents/QuickGameForm/QuickGameForm.tsx
```diff
@@ -111,6 +114,9 @@ const QuickGameForm: React.FC<IQuickGameFormProps> = ({
         const parsedInput = parseInputAsDeckData(deckLink);
         if(parsedInput.type === 'json'){
             setIsJsonDeck(true)
+            setSaveDeck(false);
+            setErrorTitle('Deck Validation Error');
+            setDeckErrorDetails('We do not support saving JSON decks at this time. Please import the deck into a deckbuilder such as SWUDB and use link import.');
```
AMMayberry1: We need to tweak the template a bit since this isn't actually a deck validation error, they're just clicking in for more details.

Instead of the below, ideally it would have maybe a yellow border and either no title at all or a title like "JSON Decks Notice" or something.

This should _not_ be the case for the decks page, since they are clicking a button to attempt to save it, so it is more correct to pop up an error message. So the version in the decks page should stay exactly the way it is.

Image

PR #531, opened 2025-10-24
Updated cosmetics

mariotee:


public/Karabast team.png
AMMayberry1: nit: can we use a dash or underscore in the filename instead of an empty space, since that can cause issues with path parsing

src/app/api/admin/cosmetics/cleanup/route.ts
```diff
@@ -0,0 +1,187 @@
...
+                },
+                { status: 500 }
+            );
+        }
+
+        const { searchParams } = new URL(request.url);
+        const action = searchParams.get('action');
+        const cosmeticId = searchParams.get('id');
+
+        switch (action) {
```
AMMayberry1: Definitely would be good to have an enum for this value

src/app/api/admin/cosmetics/cleanup/route.ts
```diff
@@ -0,0 +1,187 @@
...
+import { withAdminAuth } from '@/app/_utils/AdminAuth';
+
+// Only allow cleanup operations in development
+function checkDevelopmentMode() {
+    if (process.env.NODE_ENV !== 'development') {
+        throw new Error('Cleanup operations are only allowed in development mode');
+    }
+}
+
+export const DELETE = withAdminAuth(async (request: NextRequest) => {
```
AMMayberry1: Feels like we shouldn't be doing admin auth if these are only for local dev mode, it'd just be an extra hoop to jump through

src/app/_services/DynamoDBService.ts
```diff
@@ -0,0 +1,199 @@
...
+} from '@aws-sdk/lib-dynamodb';
+import { CosmeticOption } from '../_components/_sharedcomponents/Preferences/Preferences.types';
+
+// Singleton pattern for DynamoDB client
+let dynamoDbService: DynamoDBService;
+
+/**
+ * Get a properly initialized DynamoDB service for Next.js
+ */
+export async function getDynamoDbServiceAsync(): Promise<DynamoDBService | null> {
```
AMMayberry1: This would be good to discuss in chat, I'm still unclear what the benefits are of doing this here vs the BE serve. Since either way it's still a server call and now we have to keep two copies of these DB interfaces in sync.

I'm confident that's going to cause issues where we failed to update one side or the other, so we should either find a way to centralize the definitions or have a very clear case for why the upside is worth it

src/app/_utils/AdminAuth.ts
```diff
@@ -0,0 +1,81 @@
+import { NextRequest, NextResponse } from 'next/server';
+import { getServerSession } from 'next-auth';
+import { authOptions } from '@/app/_utils/auth';
+
+// List of allowed usernames for admin access - must match exactly what NextAuth provides
+const allowedAdminUsers = [
```
AMMayberry1: Discussed in chat already but my strong preference would be a DB document that lists these out. Also, we don't enforce unique usernames in the system, so we definitely need to be using guids instead

AMMayberry1: I think it would be good to have a few different categories of "admin" user so we can flesh out slightly different roles and permissions going forward:

- developer: access to sensitive server actions for maintenance
- moderator: analogous to what you have now, can do general player operations like interacting with customizations and eventually we can bring the player moderation actions over here
- admin: this would be for a couple "full admin" users who can do things like set add roles to other user accounts

src/app/_utils/s3Utils.ts
```diff
@@ -18,8 +18,8 @@ export const s3ImageURL = (path: string) => {
     return s3Bucket + path;
 };
 
-export const s3CardImageURL = (card: ICardData | IServerCardData | ISetCode | IPreviewCard, cardStyle: CardStyle | LeaderBaseCardStyle = CardStyle.Plain ) => {
-    if (((isGameCard(card) || isSetCodeCard(card) || isPreviewCard(card)) && !card?.setId) && !card?.id) return s3ImageURL('game/swu-cardback.webp');
+export const s3CardImageURL = (card: ICardData | IServerCardData | ISetCode | IPreviewCard, cardStyle: CardStyle | LeaderBaseCardStyle = CardStyle.Plain, cardback?: string ) => {
+    if (((isGameCard(card) || isSetCodeCard(card) || isPreviewCard(card)) && !card?.setId) && !card?.id) return cardback ? cardback : s3ImageURL('game/swu-cardback.webp');
```
AMMayberry1: nit: could we expand this out to a full function at this point, it's pretty much impossible to read

src/app/_utils/ServerAndLocalStorageUtils.ts
```diff
@@ -640,13 +661,18 @@ export const loadPreferencesFromLocalStorage = (): IPreferences => {
...
             return {
-                cardback: preferences.cardback || undefined,
                 sound: {
                     muteAllSound: preferences.sound?.muteAllSound ?? false,
                     muteCardAndButtonClickSound: preferences.sound?.muteCardAndButtonClickSound ?? false,
                     muteYourTurn: preferences.sound?.muteYourTurn ?? false,
                     muteChatSound: preferences.sound?.muteChatSound ?? false,
                     muteOpponentFoundSound: preferences.sound?.muteOpponentFoundSound ?? false,
+                },
+                cosmetics: {
```
AMMayberry1: Would be good to discuss with the whole team whether anon users should get access to customization. Our philosophy is that we want the anon category to be available for basic play but we do plan to intentionally silo certain features off for only logged-in users to incentivize people to switch over. We haven't explicitly discussed if customization would fall into that category though

src/app/_services/DynamoDBService.ts
```diff
@@ -0,0 +1,199 @@
...
+            });
+
+            return this.client?.send(command) ?? Promise.reject('DynamoDB client not initialized');
+        }, 'DynamoDB queryItems error');
+    }
+
+    // Cosmetics-specific methods
+    public async getCosmeticsAsync(): Promise<CosmeticOption[]> {
+        return this.executeDbOperationAsync(async () => {
+            const result = await this.queryItemsAsync('COSMETICS', { beginsWith: 'ITEM#' });
```
AMMayberry1: We definitely want to add some indexes (GSIs) to the dynamodb to make these queries efficient, if you can come up with a list of what indexes to add I can work on it. also they can be tested with the local db, we already have some logic on startup that sets up a basic GSI and we'll want to add any additional ones to that logic anyway for local dev purposes

src/app/api/admin/cosmetics/cleanup/route.ts
mariotee: oh yeah i can actually remove these and the button in the mod page for non-dev env

public/Karabast team.png
mariotee: @CheBato do we still need this?

src/app/_services/DynamoDBService.ts
mariotee: yeah we can probably just add one for now on type. and then re-visit if we consider Patreon enabled card backs to filter those out easier

src/app/_utils/s3Utils.ts
mariotee: should be a little more readable now

src/app/_utils/AdminAuth.ts
mariotee: agreed. we can flesh that out at another time. for now, i see probably only a few people to be adding CBs if any. i can take that burden if people start bombarding us with requests after this is release

src/app/_services/ServerApiService.ts
AMMayberry1: Is this file still needed now that the cosmetics DB stuff was migrated to the BE repo?

src/app/_contexts/CosmeticsContext.tsx
```diff
@@ -0,0 +1,89 @@
...
+        }
+        return cosmetics.playmats.find((pm) => pm.id === id) || getCosmeticDefault(RegisteredCosmeticType.Playmat);
+    }
+
+    React.useEffect(() => {
+        const fetchCosmetics = async () => {
+            const baseUrl = process.env.NODE_ENV === 'development'
+                ? 'http://localhost:3000'
+                : 'https://karabast.net';
+            const response = await fetch(`${baseUrl}/api/cosmetics`);
```
AMMayberry1: Looks like we can just omit the `baseUrl` and it'll work correctly, see for example our usage of the `swudbdeck` route in `fetchDeckData.ts`

src/app/_temp/fallback-cosmetics.json
AMMayberry1: Can this be removed now that we're handling this on the BE side?

src/app/_utils/AdminAuth.ts
```diff
@@ -0,0 +1,51 @@
+// Admin authentication middleware
+import { NextRequest, NextResponse } from 'next/server';
+import { getServerApiService } from '@/app/_services/ServerApiService';
+
+export function withAdminAuth(
+    role: string,
```
AMMayberry1: We should use the enum type here as well for consistency

src/app/api/admin/cosmetics/cleanup/route.ts
```diff
@@ -0,0 +1,181 @@
...
+}
+
+enum CleanupAction {
+    Single = 'single',
+    All = 'all',
+    Reset = 'reset'
+}
+
+// DELETE endpoint to handle cleanup operations
+export const DELETE = withAdminAuth(AdminRole.Developer, async (request: NextRequest) => {
```
AMMayberry1: Same note as on the BE side - I'm fine with this being available at the moderator level

src/app/api/admin/cosmetics/cleanup/route.ts
```diff
@@ -0,0 +1,181 @@
...
+                                await s3Service.deleteFile(s3Key);
+                            }
+                        }
+                    } catch (s3Error) {
+                        console.warn(`Failed to delete S3 file for cosmetic ${cosmeticId}:`, s3Error);
+                        // Continue with database deletion even if S3 deletion fails
+                    }
+                }
+
+                await serverApiService.deleteCosmeticAsync(cosmeticId, cookies || undefined);
```
AMMayberry1: Let's maybe do the DB deletion first just to reduce the risk of ending up with a DB entry with no matching file

src/app/api/admin/cosmetics/upload/route.ts
AMMayberry1: What's the reason for having both an `upload` and an `upload-file`?

src/app/api/admin/cosmetics/route.ts
```diff
@@ -0,0 +1,133 @@
...
+            {
+                error: 'Error fetching cosmetics from server',
+                details: error instanceof Error ? error.message : 'Unknown error'
+            },
+            { status: 500 }
+        );
+    }
+};
+
+export const POST = withAdminAuth(AdminRole.Moderator, async (_request: NextRequest) => {
```
AMMayberry1: Having trouble understanding this one - what would `POST /api/admin/cosmetics` be used for? mass upload?

src/app/api/cosmetics/route.ts
```diff
@@ -0,0 +1,15 @@
+import { NextResponse } from 'next/server';
+import { fetchCosmeticsDataAsync } from '@/app/_utils/fetchCosmeticsData';
+
+export async function GET() {
```
AMMayberry1: What's the difference between this and the admin version?

src/app/GameBoard/page.tsx
```diff
@@ -12,16 +12,25 @@ import PopupShell from '../_components/_sharedcomponents/Popup/Popup';
...
     const [sidebarOpen, setSidebarOpen] = useState(sidebarState);
     const [isPreferenceOpen, setPreferenceOpen] = useState(false);
     const [userClosedWinScreen, setUserClosedWinScreen] = useState(false);
+    const user = gameState?.players[connectedPlayer].user;
+    const background = getBackground(user?.cosmetics?.background ?? null);
+    const playMatsDisabled = user?.cosmetics?.disablePlaymats ?? false;
+    const myPlaymatId = !playMatsDisabled ? user?.cosmetics?.playmat : 'none';
+    const myPlaymat = myPlaymatId && myPlaymatId !== 'none' ? getPlaymat(myPlaymatId) : null;
+    const opponentUser = gameState?.players[getOpponent(connectedPlayer)].user;
+    const theirPlaymatId = !playMatsDisabled ? opponentUser?.cosmetics?.playmat : 'none';
```
AMMayberry1: Is there a reason to have a specific `'none'` value for these instead of just using `null`?

src/app/Preferences/page.tsx
```diff
@@ -47,7 +48,7 @@ const Preferences: React.FC = () => {
             <PreferencesComponent
                 sidebarOpen={false}
                 isPreferenceOpen={true}
-                tabs={['general','soundOptions']}
+                tabs={['general','soundOptions'].concat(useUser().user ? ['cosmetics'] : [])}
```
AMMayberry1: In case we end up needing it elsewhere in here, can we unpack `user` at the top of the file with `const { user } = userUser()` like we do in other cases

src/app/_components/Gameboard/OpponentCardTray/OpponentCardTray.tsx
```diff
@@ -36,35 +36,36 @@ const OpponentCardTray: React.FC<IOpponentCardTrayProps> = ({ trayPlayer, prefer
     const phase = gameState.phase;
     const warning = gameState?.players[connectedPlayer]?.timeRemainingStatus === 'Warning';
     const danger = gameState?.players[connectedPlayer]?.timeRemainingStatus === 'Danger';
+    const opponentsCardback = gameState?.players[getOpponent(connectedPlayer)].user?.cosmetics?.cardback;
```
AMMayberry1: Just a general note that I did a couple quick checks and it looks like the customizations work a bit inconsistently for spectators. Our design generally inherits the game board view of "player 1" for spectation, but for this stuff it means that (a) the spectator's own settings for things like background and playmats enabled / disabled aren't respected and (b) the player 1 card backs aren't visible.

that's enough of a secondary issue that I'd be fine to punt on that for a future PR if we just want to fully disable customizations when spectating for now

src/app/_services/S3Service.ts
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            const credentials = {
+                accessKeyId: process.env.AWS_ACCESS_KEY_ID,
+                secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
+                sessionToken: undefined as string | undefined
+            };
+
+            // Add session token if available (for temporary credentials)
+            if (process.env.AWS_SESSION_TOKEN) {
```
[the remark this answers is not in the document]
mariotee: i had to add that for the creds it was given to interface with S3. my key and secret were not enough

src/app/_temp/fallback-cosmetics.json
mariotee: we'll remove both yes. since it'll be assumed to be initialized

src/app/_services/S3Service.ts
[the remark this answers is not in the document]
mariotee: i* was given

src/app/GameBoard/page.tsx
```diff
@@ -53,17 +62,19 @@ const GameBoard = () => {
     // const winners = ['order66']
     // we set tabs
     // ['endGame','keyboardShortcuts','cardSleeves','gameOptions']
+    const anonymousPattern = /anonymous [0-9a-f]{6}/;
```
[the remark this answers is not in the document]
mariotee: ended up not needing this

src/app/GameBoard/page.tsx
CheBato: good question. Resolved

src/app/api/cosmetics/route.ts
CheBato: i have yet to find a good reason for it. Might just delete the file since i don't see it being used anywhere.

src/app/api/admin/cosmetics/route.ts
CheBato: i have yet to find a good reason for it. Might just delete the file since i don't see it being used anywhere.

src/app/api/admin/cosmetics/route.ts
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        try {
+            const cosmetics = await serverApiService.getCosmeticsAsync();
+
+            return NextResponse.json(
+                {
+                    success: true,
+                    cosmetics,
+                    count: cosmetics.length,
```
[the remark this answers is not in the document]
CheBato: i have yet to find a good reason for it. Might just delete the file since i don't see it being used anywhere.

src/app/mod/ModPageClient.tsx
```diff
@@ -0,0 +1,940 @@
...
+                    <StyledTextField
+                        label="Title"
+                        value={cosmeticTitle}
+                        onChange={(e) => setCosmeticTitle(e.target.value)}
+                        disabled={uploadLoading}
+                        fullWidth
+                    />
+
+                    <Box sx={{ display: 'flex', gap: 1, alignItems: 'flex-start' }}>
+                        <StyledTextField
```
AMMayberry1: I don't think there's any advantage to making this configurable, we should just generate a guid under the hood when image is uploaded

src/app/mod/ModPageClient.tsx
```diff
@@ -0,0 +1,949 @@
...
+                                                            {cosmetic.type} • {cosmetic.id}
+                                                        </Typography>
+                                                        <Button
+                                                            size="small"
+                                                            color="error"
+                                                            variant="outlined"
+                                                            onClick={() => handleDeleteSingleCosmetic(cosmetic.id)}
+                                                            sx={{ mt: 1, fontSize: '0.7rem' }}
+                                                        >
+                                                            Delete
```
AMMayberry1: We need to make sure the Delete button isn't available for the default images

src/app/mod/ModPageClient.tsx
```diff
@@ -0,0 +1,949 @@
...
+    const [uploadError, setUploadError] = useState<string | null>(null);
+    const [uploadSuccess, setUploadSuccess] = useState(false);
+
+    // Cleanup dialog states
+    const [cleanupDialogOpen, setCleanupDialogOpen] = useState(false);
+    const [cleanupLoading, setCleanupLoading] = useState(false);
+    const [cleanupError, setCleanupError] = useState<string | null>(null);
+    const [cleanupSuccess, setCleanupSuccess] = useState(false);
+
+    // Upload form states
```
AMMayberry1: Not sure exactly where to put this comment - this is not a hard blocker, but just a note that this double-scroll situation is a bit weird. If there's an easy way to separate them visually and make them not-overlapping that'd be ideal, but it can definitely be a thing we solve after this PR is done:

Image

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/CosmeticsTab.tsx
```diff
@@ -0,0 +1,362 @@
...
+    const [disablePlaymats, setDisablePlaymats] = useState<boolean>(false);
+    const [expandedAccordion, setExpandedAccordion] = useState<string>('cardbacks'); // Default to cardbacks expanded
+
+
+    useEffect(() => {
+        if (user) {
+            const { cosmetics } = user.preferences;
+            if (cosmetics) {
+                setSelectedCardback(cosmetics.cardback ?? 'Default');
+                setSelectedBackground(cosmetics.background ?? 'Default');
```
AMMayberry1: Got this error when I tried to join a game with a new logged-in user and default settings for everything: 

Image

src/app/mod/ModPageClient.tsx
```diff
@@ -0,0 +1,933 @@
...
+                                        },
+                                    }}
+                                />
+                            }
+                            label={
+                                <Box>
+                                    <Typography variant="body2" color="white">
+                                        Apply darkening effect
+                                    </Typography>
+                                    <Typography variant="caption" color="textSecondary">
```
AMMayberry1: This text is also dark

Image

PR #542, opened 2025-11-03
Disable Next Set Preview format

AMMayberry1:


moyerr (approved):
Nice! I like the announcement update too

PR #543, opened 2025-11-05
feat: add my-swu.com to deck sources

Akryum:


src/app/DeckPage/page.tsx
```diff
@@ -434,6 +436,15 @@ const DeckPage: React.FC = () => {
             },
             boxShadow: '0 0 5px #4CFF85',
         },
+        mySwuTag: {
+            borderColor: '#F6BD26',
```
AMMayberry1: This is really close to the color for swustats, do you mind picking something a little more distinct from it?

PR #545, opened 2025-11-07
Render promptState choices as tray buttons if player has selected cards

Ravarix:
In order to increase readability this PR renders prompt choices as tray buttons instead of in the popup modal.

Currently this is also limited to if there are exactly 2 choices, as that's the current tray button UI limit.

image


src/app/_contexts/Game.context.tsx
```diff
@@ -102,6 +102,17 @@ export const GameProvider = ({ children }: { children: ReactNode }) => {
                 playSound('yourTurn');
             }
 
+
+            let hasSelectedCards = false;
```
danbastin: Can we actually start up a `gamestateHelpers` file in `/app/_utils`? I think thers some other stuff like this that could be good to move out and have available broadly.

PR #553, opened 2025-11-22
Update lobby to use newer deck selection components

AMMayberry1:
image


src/app/_hooks/useDeckManagement.ts
```diff
@@ -0,0 +1,143 @@
...
+    fetchDecks: () => Promise<void>;
+}
+
+export const useDeckManagement = (): IDeckManagementState => {
+    const { user } = useUser();
+    const { data: session } = useSession();
+    
+    // Deck Preferences State
+    const [showSavedDecks, setShowSavedDecks] = useState<boolean>(() => {
+        if (typeof window === 'undefined') return true;
```
danbastin: I like getting this stuff into a hook. Can we clean out the `typeof window` stuff? Otherwise I think this looks good.

PR #554, opened 2025-11-24
implemented UI for animation system

mariotee:
feat: added defeat animations for units and upgrades

feat: added animation for shield breaks

chore: prep for animations prefs

feat: added animations tab

feat: added healing animation

# Conflicts:
#	src/app/_components/_sharedcomponents/Cards/GameCard.tsx

### DIFF
--- a/src/animations/animations.css
+++ b/src/animations/animations.css
@@ -1,0 +1,381 @@
+/* Damage Flash Overlay */
+.damage-flash {
+  position: absolute;
+  top: 0;
+  left: -2px;
+  width: 100%;
+  height: calc(100% - 8px);
+  padding: 0 2px;
+  border-radius: 12px;
+  background-color: rgba(220, 38, 38, 0.5);
+  z-index: 1000;
+  pointer-events: none;
+  animation: damageFlash 0.3s ease-in-out forwards;
+}
+
+@keyframes damageFlash {
+  0% {
+    opacity: 0;
+  }
+  15% {
+    opacity: 1;
+  }
+  75% {
+    opacity: 1;
+  }
+  100% {
+    opacity: 0;
+  }
+}
+
+/* Damage Numeric Label */
+.damage-label {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-size: 36px;
+  font-weight: 600;
+  color: white;
+  text-shadow: 1px 1px 0px rgba(0, 0, 0, 0.60);
+  z-index: 1001;
+  pointer-events: none;
+  animation: damageSlideUp 0.3s ease-out forwards;
+}
+
+@keyframes damageSlideUp {
+  0% {
+    transform: translateY(0);
+    opacity: 1;
+  }
+  60% {
+    transform: translateY(-20px);
+    opacity: 1;
+  }
+  100% {
+    transform: translateY(-40px);
+    opacity: 0;
+  }
+}
+
+/* Heal Flash Overlay */
+.heal-flash {
+  position: absolute;
+  top: 0;
+  left: -2px;
+  width: 100%;
+  height: calc(100% - 8px);
+  padding: 0 2px;
+  border-radius: 12px;
+  background-color: rgba(0, 186, 255, 0.5);
+  z-index: 1000;
+  pointer-events: none;
+  animation: healFlash 0.3s ease-in-out forwards;
+}
+
+@keyframes healFlash {
+  0% {
+    opacity: 0;
+  }
+  15% {
+    opacity: 1;
+  }
+  75% {
+    opacity: 1;
+  }
+  100% {
+    opacity: 0;
+  }
+}
+
+/* Heal Numeric Label */
+.heal-label {
+  position: absolute;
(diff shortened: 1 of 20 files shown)


src/app/_contexts/Game.context.tsx
```diff
@@ -235,21 +239,56 @@ export const GameProvider = ({ children }: { children: ReactNode }) => {
...
+
+            const hasShieldAnimations = gameState.animations?.events?.some(
+                (event: any) => event.type === 'LOSE_SHIELD' || event.type === 'GAIN_SHIELD'
+            );
+
+            // Process animations BEFORE updating game state if there are defeat or shield animations
+            // (these animations need the elements to still be in DOM)
+            if ((hasDefeatAnimations || hasShieldAnimations) && gameState.animations) {
+                try {
+                    await processAnimations(gameState.animations);
```
danbastin: I like the idea of making this function async and just awaiting the animation instead of trying to juggle pre and post animation game states so thats nice. 

I think what might be better here though is to have a property on events that mark them as pre or post update and then disambiguate here. That way the BE can just tell us where in the timeline each event should be rendered. We might also need to tell the manager which animations to process in each window. I'm not sure if we might ahve some events that would require an update before AND after update. Perhaps on a `When defeated` trigger?

src/app/_components/_sharedcomponents/Cards/GameCard.tsx
```diff
@@ -582,15 +582,14 @@ const GameCard: React.FC<IGameCardProps> = ({
             <Box
                 sx={styles.card}
                 onClick={handleClick}
+                onMouseEnter={handlePreviewOpen}
```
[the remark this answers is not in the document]
mariotee: i see.. and i needed them up a layer due to the defeat animation showing the black box behind it. i'll revert this and then put some styling to make that box transparent

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/AnimationOptionsTab.tsx
```diff
@@ -0,0 +1,343 @@
...
+    const [animationMode, setAnimationMode] = useState<AnimationMode>(() => {
+        if (user && user?.preferences?.animations) {
+            return getAnimationModeFromPreferences(user.preferences.animations);
+        } else {
+            const localPreferences = loadPreferencesFromLocalStorage();
+            return getAnimationModeFromPreferences(localPreferences.animations);
+        }
+    });
+
+    const [fastAnimations, setFastAnimations] = useState<boolean>(() => {
```
ejmudrak: I can't tell if animation speed being a boolean makes this more clean and readable, or a little confusing.

src/animations/animations.css
```diff
@@ -0,0 +1,381 @@
...
+@keyframes shieldFadeOut {
+  0% {
+    opacity: 1;
+  }
+  100% {
+    opacity: 0;
+  }
+}
+
+/* Top left piece */
```
ejmudrak: The shrapnel shield defeat looks good! I think CSS animation code here makes sense too.

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/AnimationOptionsTab.tsx
```diff
@@ -0,0 +1,343 @@
...
+  }
+  50% {
+    border-color: rgba(0, 180, 255, 0.6);
+    box-shadow: 0 0 8px rgba(0, 180, 255, 0.6);
+  }
+  100% {
+    border-color: rgba(0, 140, 255, 0.4);
+    box-shadow: 0 0 4px rgba(0, 140, 255, 0.4);
+  }
+`;
```
ejmudrak: Are there any shared/global CSS or animations defined in this codebase? I see that this `pulseBorder` animation's code is duplicated for each settings page at least.

Standardizing and having a source of truth for styles is something that I'd be willing to take on sometime.

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/AnimationOptionsTab.tsx
```diff
@@ -0,0 +1,343 @@
...
+                            }
+                            sx={styles.radioOption}
+                        />
+                        <FormControlLabel
+                            value="onlyDamage"
+                            control={<Radio />}
+                            label={
+                                <Box>
+                                    <Typography variant="body1" fontWeight="bold">Show Only Damage Animations</Typography>
+                                    <Typography variant="body2" color="text.secondary">
```
ejmudrak: These subtitle Typography components need to have a different color, they're currently defaulting to black and indistinguishable from the background

ejmudrak: color: '#878787' is what the other tabs are using. The theme could be updated to use that color as `text.secondary` too.

ejmudrak (commented):
I'm new to this codebase, but some really exciting changes here. Especially for new players (like myself) that may feel like "wait wtf just happened" when an action is made against one of my units. Good stuff here!

PR #654, opened 2026-03-12
Chebato/mod tools

CheBato:
User Management allows moderators to search for players by username or ID, view their account details and moderation history, and apply actions: Mute (timed or permanent, preventing chat usage), Warning (logged note on the player's record), and Rename (forces a username change). Active actions can be cancelled, and all actions are tracked with timestamps, duration, and the issuing moderator.

**Auto-cleanup:** The ModActionCache extends TimedCache and refreshes once every 24 hours in production. On each refresh it re-fetches all active mod actions from DynamoDB. Any action whose expiresAt is in the past gets dropped from the in-memory cache, silently un-muting the user on their next refresh.

Testing the cache cleanup locally:

1. Start the local DynamoDB docker container (docker compose up dynamodb)
2. Run the seedAdmin script to activate yourself as a moderator
3. Set REFRESH_INTERVAL_MINUTES=5 in modActionCache
4. In the seedModAction script, set MY_USER_ID = <your_user_id> and EXPIRES_IN_MINUTES = 2, then run it
5. [optional] Set the ENV variable on the BE DEBUG_MOD_ACTION_CACHE=true
6. Start the server. The seeded mute will be active. After 2 minutes it expires, and the next cache refresh (within 1 min) 
7. cleans it up. Confirm on the Mod Page that the action shows as "Expired".

BE PR: https://github.com/SWU-Karabast/forceteki/pull/2288


src/app/mod/subpages/UserManagementTab.tsx
```diff
@@ -0,0 +1,647 @@
...
+                                            size="small"
+                                            sx={{ minWidth: '30px', maxWidth:'150px' }}
+                                        >
+                                            <MenuItem value={DurationUnit.Days}>Days</MenuItem>
+                                            <MenuItem value={DurationUnit.Weeks}>Weeks</MenuItem>
+                                            <MenuItem value={DurationUnit.Permanent}>Permanent</MenuItem>
+                                        </StyledTextField>
+
+                                        {durationUnit !== DurationUnit.Permanent && (
+                                            <StyledTextField
```
AMMayberry1: Having the same word in both boxes is a little unclear, maybe we can put like "# of weeks" or something for the actual text box? Also maybe if it could be should up using the "hint" text style that would work better

Image

AMMayberry1: Actually, another idea would be to use a number field: https://mui.com/material-ui/react-number-field/

Then maybe it could be styled as <number field> <duration field>, with the number field on the left. So that way when you enter it, the text will literally show "5 weeks".

I feel like I'm explaining that poorly, lmk if it's not clear

src/app/mod/subpages/UserManagementTab.tsx
```diff
@@ -0,0 +1,647 @@
...
+
+            {/* Confirmation Dialog */}
+            <ConfirmationDialog
+                open={confirmDialog.open}
+                title={confirmDialog.title}
+                message={confirmDialog.message}
+                onConfirm={confirmDialog.onConfirm}
+                onCancel={() => setConfirmDialog((prev) => ({ ...prev, open: false }))}
+                confirmButtonText={confirmDialog.confirmText || 'Confirm Action'}
+                cancelButtonText="Cancel Action"
```
AMMayberry1: The text for this prompt is pretty confusing... since the thing we're trying to do is to "cancel" an action, it's not clear what the blue "cancel" button will do.

Maybe they should say something like "Do not cancel" and "Proceed with cancel", I feel like that's still a bit clunky though...

Image

src/app/_utils/ModerationUtils.ts
```diff
@@ -28,4 +39,46 @@ export const checkIfModerationExpired = (moderation: IModerationAction, updateMo
     if (timeDiff <= 0){
         updateModerationSeenStatus(null);
     }
-}
\ No newline at end of file
+}
+
+export const formatDate = (dateStr: string) => {
+    const date = new Date(dateStr);
+    return date.toLocaleDateString('en-GB', { day: '2-digit', month: '2-digit', year: 'numeric', hour: '2-digit', minute: '2-digit' });
```
AMMayberry1: For some reason these are coming out as DD/MM/YYYY? Was expecting MM/DD/YYYY

Image

src/app/mod/subpages/UserManagementTab.tsx
```diff
@@ -0,0 +1,647 @@
...
+                                                            </Typography>
+                                                            <Typography sx={{ color: '#B0B0B0', fontSize: '0.75em' }}>
+                                                                {action.note}
+                                                            </Typography>
+                                                        </Box>
+                                                    )}
+                                                    <Box sx={{ display: 'flex', flexWrap: 'wrap', gap: '1rem' }}>
+                                                        <Box>
+                                                            <Typography sx={{ color: '#8C8C8C', fontSize: '0.75rem', mb:'0px' }}>
+                                                                Moderator ID:
```
AMMayberry1: How much work would it be to also get the moderator's username? In addition to the ID. The accountability is important, but no one will remember any of the IDs. They can be looked up but that's a headache since it means closing out the user profile you're currently looking at

src/app/mod/subpages/UserManagementTab.tsx
```diff
@@ -0,0 +1,647 @@
...
+                        <Box>
+                            <Typography sx={styles.sectionTitle}>User details</Typography>
+                            <Box sx={{ display: 'flex', flexDirection: 'column', gap: '0.35rem' }}>
+                                <Box sx={styles.userDetailRow}>
+                                    <Typography sx={styles.userDetailLabel}>Username:</Typography>
+                                    <Typography sx={styles.userDetailValue}>{selectedPlayer.username}</Typography>
+                                </Box>
+                                <Box sx={styles.userDetailRow}>
+                                    <Typography sx={styles.userDetailLabel}>Player ID:</Typography>
+                                    <Typography sx={{ ...styles.userDetailValue, fontSize: '0.75rem', color: '#B0B0B0' }}>
```
AMMayberry1: Is it intentional for this to have a different size and color than the surrounding text?

Image

src/app/mod/subpages/UserManagementTab.tsx
```diff
@@ -0,0 +1,647 @@
...
+                                    text={submitLoading ? '' : actionType}
+                                    buttonFnc={handleSubmitAction}
+                                    disabled={!canSubmit || submitLoading}
+                                />
+                            </Box>
+
+                            {/* Notes */}
+                            <Typography sx={{ ...styles.sectionTitle, mt: '0.75rem' }}>Moderator notes for action</Typography>
+                            <StyledTextField
+                                placeholder="Notes about the action"
```
AMMayberry1: We should probably make this required in order to submit any action, except maybe in the case of a rename

src/app/mod/subpages/UserManagementTab.tsx
```diff
@@ -0,0 +1,647 @@
...
+                                <Box sx={styles.userDetailRow}>
+                                    <Typography sx={styles.userDetailLabel}>Created:</Typography>
+                                    <Typography sx={styles.userDetailValue}>
+                                        {formatDate(selectedPlayer.createdAt)}
+                                    </Typography>
+                                </Box>
+                                {selectedPlayer.isMuted && (
+                                    <Box sx={styles.userDetailRow}>
+                                        <Typography sx={{ color: '#ef5350', fontSize: '0.875rem', fontWeight: 600 }}>
+                                            Currently muted
```
AMMayberry1: I like this but it maybe deserves its own text section so people can find it easier. That will also help if there are multiple stacked together. Maybe just a box here named "active mod actions" or something?

src/app/mod/subpages/UserManagementTab.tsx
```diff
@@ -0,0 +1,647 @@
...
+                    {successMessage}
+                </Alert>
+            )}
+
+            <Box sx={styles.container}>
+                {/* ==================== Left Column ==================== */}
+                <Box sx={styles.leftColumn}>
+                    {/* Search */}
+                    <Box>
+                        <Typography sx={styles.sectionTitle}>Find user</Typography>
```
AMMayberry1: This is the nittiest of nits, but could we do Every Word Capitalization on these section titles? Just triggering my OCD, lol...

src/app/mod/subpages/CosmeticsManagerTab.tsx
AMMayberry1, line 1: nit on the filename: "tab" should be capitalized separately

src/app/_utils/ModerationUtils.ts
```diff
@@ -28,4 +39,46 @@ export const checkIfModerationExpired = (moderation: IModerationAction, updateMo
...
+
+export const getActionLabel = (action: IModActionResponse): string => {
+    const date = formatDate(action.createdAt);
+    switch (action.actionType) {
+        case ModActionType.Mute:
+            return `${date} Muted (${formatDuration(action.durationDays)})`;
+        case ModActionType.Warning:
+            return `${date} Warning`;
+        case ModActionType.Rename:
+            return `${date} Renamed`;
```
AMMayberry1, line 59: Since the rename operation also has a "pending" status we should show that in the action status, currently it just shows this even if a rename is pending:

Image

src/app/_utils/ModerationUtils.ts
```diff
@@ -1,18 +1,29 @@
 import { IModerationAction, IUser } from '@/app/_contexts/UserTypes';
+import {
+    IModActionResponse, ModActionType
+} from "@/app/_components/_sharedcomponents/Preferences/Preferences.types";
+
+const PERMANENT_DURATION_DAYS = 36500;
```
AMMayberry1: This is a bit too much of a hack for my taste, if we want to add support for a permanent mute then we should add a dedicated option for it imo. It makes things cleaner internally, and also it gets around weird messages like the below.

My vote would be to remove or comment it out and add in a followup PR

Image

src/app/_utils/ModerationUtils.ts
```diff
@@ -28,4 +39,46 @@ export const checkIfModerationExpired = (moderation: IModerationAction, updateMo
...
+    if (durationDays >= PERMANENT_DURATION_DAYS/2) return 'Permanent';
+    if (durationDays % 7 === 0 && durationDays >= 7) return `${durationDays / 7} week${durationDays / 7 !== 1 ? 's' : ''}`;
+    return `${durationDays} day${durationDays !== 1 ? 's' : ''}`;
+};
+
+export const getActionLabel = (action: IModActionResponse): string => {
+    const date = formatDate(action.createdAt);
+    switch (action.actionType) {
+        case ModActionType.Mute:
+            return `${date} Muted (${formatDuration(action.durationDays)})`;
```
AMMayberry1, line 55: Let's also add the duration as a property within the tile itself, it should be findable somewhere other than the title string

AMMayberry1:
One additional thing I noticed - I gave user a rename and a mute simultaneously, and when I next logged in with that account I saw the rename popup but not the mute one. Not sure if that's a pre-existing issue

src/app/mod/subpages/UserManagementTab.tsx
CheBato: yessir will change. I think the original reasoning for me was so you can immediately find it

src/app/mod/subpages/UserManagementTab.tsx
CheBato: I'll get it done.

CheBato: Not much work should be pretty simple.

src/app/mod/subpages/UserManagementTab.tsx
```diff
@@ -0,0 +1,669 @@
...
+                                    </Typography>
+                                </Box>
+                                <Box sx={styles.userDetailRow}>
+                                    <Typography sx={styles.userDetailLabel}>Created:</Typography>
+                                    <Typography sx={styles.userDetailValue}>
+                                        {formatDate(selectedPlayer.createdAt)}
+                                    </Typography>
+                                </Box>
+                                <Box>
+                                    <Typography sx={styles.sectionTitle}>Active mod actions:</Typography>
```
AMMayberry1, line 419: nit: would be good to add a little bit of context when this is empty, just looks like the below currently. E.g., we could remove the box and just say "No active mod actions" or even keep the box and just add the word "None" inside of it

Image

src/app/mod/subpages/UserManagementTab.tsx
```diff
@@ -0,0 +1,669 @@
...
+                                    <MenuItem value={ModActionType.Rename}>Force Rename</MenuItem>
+                                    <MenuItem value={ModActionType.Mute}>Mute</MenuItem>
+                                </StyledTextField>
+
+                                {/* Duration - only for Mute */}
+                                {actionType === ModActionType.Mute && (
+                                    <>
+                                        <StyledTextField
+                                            type="number"
+                                            label="amount"
```
AMMayberry1: nit: should be capitalized as "Amount". also, the word "Duration" might be slightly clearer what the box is for imo

AMMayberry1: Could we switch this to being a placeholder instead of a label? Right now it looks like the box is pre-filled with the word "amount" which might make the intent confusing

AMMayberry1 (approved):

Issue #556, opened 2025-11-27
Enable chat disabling for logged-in users (UI and backend integration)


danbastin:
**Feature Request:** Add the ability for logged-in users to disable chat, with UI and backend integration. This feature should mirror the behavior currently implemented for anonymous users.

**Requirements:**
- Implement UI controls to allow logged-in users to toggle chat functionality on or off.
- Ensure that the backend properly recognizes and handles this setting for logged-in users, preventing them from sending/receiving chat messages as appropriate.
- Update the chat box UI to display a notification indicating when chat is disabled for the current user.
- If a user has chat disabled, ensure that chat is disabled for the opponent as well and a message displays with the reason so they are aware that the other user cannot see messages. (should function similar to chat blocking for anon users)

**Acceptance Criteria:**
- UI toggle present and functioning for logged-in users.
- Chat box shows appropriate notification for both self and opponent when chat is disabled.

---
*Further discussion may be required to finalize the UI/notification design details.*

PR #657, opened 2026-03-18
MuteOption

CheBato:
#556 

Testing suite
- [x] If user is Anon he shouldn't have this option
- [x] If user is Logged in, should have option and able to disable chat (check lobby, quickmatch)
- [x] Display correct message if both users have chat disabled. 
- [x] If user is Muted should display muted.
- [x] If user logs out and logs back in the mute option should still function 

BE PR: https://github.com/SWU-Karabast/forceteki/pull/2297


src/app/_utils/genericPreferenceFunctions.ts
```diff
@@ -104,5 +116,8 @@ const getDefaultPreferences = (): IPreferences => ({
         background: undefined,
         // playmat: undefined,
         // disablePlaymats: false,
+    },
+    gameOptions: {
+        muteChat: false,
```
AMMayberry1, line 121: I think it would also be useful to add a hidden user property that lets us override the `muteChat` option so that it's always enabled. We've already had at least one case where a player self-identified as a child and we gave them an extended mute, it would be good if we can have a mod option to just lock chat for such accounts.

In that case the mute option in Preferences will be selected and disabled so that they can't change it, probably with some yellow explainer text indicating why it is that way. Let me know if that makes sense!

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GameOptionsTab.tsx
```diff
@@ -0,0 +1,87 @@
...
+                    />
+                </Box>
+                {!user && (
+                    <Typography sx={{ ml: '1rem', color: '#888', fontSize: '0.85rem' }}>
+                        Log in to save game options.
+                    </Typography>
+                )}
+                {saveStatus === 'success' && (
+                    <Alert severity="success" sx={{ ...styles.statusContainer, background: 'none', color: 'green' }}>
+                        Preference saved successfully.
```
AMMayberry1: Feels like we should have a "Save Changes" button in this tab as well for the sake of consistency with the Sound Options tab

src/app/_components/_sharedcomponents/Chat/Chat.tsx
```diff
@@ -70,25 +70,30 @@ const Chat: React.FC<IChatProps> = ({
...
             case didCurrentUserMuteChat:
                 return {
                     reason: ChatDisabledReason.UserDisabledChat,
                     message: 'You disabled chat',
                     borderColor: 'yellow'
                 };
+            case doesUserHaveChatMuted:
+                return {
+                    reason: ChatDisabledReason.UserDisabledChat,
+                    message: 'You have chat disabled',
```
AMMayberry1: nit: maybe for slightly added clarity, `'You have chat disabled in your account preferences'`

src/app/_utils/genericPreferenceFunctions.ts
CheBato: that makes complete sense. I would recommend we do this after we merge the ModActions PR so we can just cleanly add it afterwards. Since the flow will be cleaner.

src/app/_components/_sharedcomponents/Chat/Chat.tsx
AMMayberry1: Currently this gets past the "disable chat" restriction, which I think it shouldn't. Can we make it so that setting will also disable chat for private games

CheBato: sure thing!

moyerr (approved):

PR #661, opened 2026-03-24
Disclose Mock Delay UI

mnechromancer:
Adding a timer feature to the Disclose mechanic: Opponents have an unfair advantage if a player cannot Disclose because the client simply skips the popup, hinting to the opponent that the player lacks the cards for Disclose. With this PR, a popup will still appear and the "skip" button will be highlighted, draining away based on a random time generated by the [backend](https://github.com/SWU-Karabast/forceteki/pull/2306).

Currently an early version, the timer is visible on the "skip" button and the popup with some notification text appears when appropriate. A preview is below.

https://github.com/user-attachments/assets/2613b3ca-5174-4b11-bea0-4710f34d7b74


moyerr:
Subsumed by #783

PR #666, opened 2026-04-17
add keyboard shortcuts for undo and pass

sethhelman:
"u" key for undo, spacebar for pass

### DIFF
--- a/src/app/_components/Gameboard/_subcomponents/PlayerTray/CardActionTray.tsx
+++ b/src/app/_components/Gameboard/_subcomponents/PlayerTray/CardActionTray.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useRef } from 'react';
 import { Button, Box } from '@mui/material';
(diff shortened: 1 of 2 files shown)


src/app/_components/Gameboard/_subcomponents/PlayerTray/CardActionTray.tsx
```diff
@@ -235,6 +235,28 @@ interface IPromptButtonProps {
...
+            if (event.code === 'Space') {
+                event.preventDefault();
+                // Only trigger if this is a pass button
+                if (button.arg === 'pass' || button.arg === 'passAbility') {
+                    passButtonRef.current?.click();
+                }
+            }
+        };
+
+        window.addEventListener('keydown', handleKeyPress);
```
CheBato, line 257: Maybe what we could do is add a if statement here checking whether its a pass button or not so that each button doesn't get a keydown event that doesn't do anything for them? Same thing for the Undo. What do you think?

CheBato:
Since we have another PR should we close this one?

PR #680, opened 2026-05-14
Add timer visibility preferences

moyerr:
Engine PR: SWU-Karabast/forceteki#2394

### Settings

| Preferences pane | In-game menu |
| :---: | :---: |
|  |  |

### Timer UI

| Standard | Hide turn timer | Hide all timers |
| :---: | :---: | :---: |
|  |  |  |

### DIFF
--- a/src/app/_components/Gameboard/_subcomponents/OpponentTray/GameTimer.tsx
+++ b/src/app/_components/Gameboard/_subcomponents/OpponentTray/GameTimer.tsx
@@ -7,6 +7,7 @@ import { useGame } from '@/app/_contexts/Game.context';
 import { useUser } from '@/app/_contexts/User.context';
 import { TimerVisibility } from '@/app/_contexts/UserTypes';
 import { DEFAULT_TIMER_VISIBILITY } from '@/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GameOptionsTab';
+import TimerWarningIcon from './TimerWarningIcon';
 
 const TIMER_STEP = 100;
 
@@ -91,7 +92,9 @@ const GameTimer: React.FC = ({ ...props }) => {
     const showOpponentMainTime = !playerIsActive && opponentIsActive && !isTurnTime;
 
     if (timerVisibility === TimerVisibility.HideAll) {
-        return null;
+        // Clock is hidden, but the warning icon still appears in this slot when main time
+        // is low so the player has a chance to react before timing out.
+        return <TimerWarningIcon />;
     }
 
     return (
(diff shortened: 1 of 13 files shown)


src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GameOptionsTab.tsx
```diff
@@ -92,28 +124,78 @@ function GameOptionsTab({ setHasNewChanges }: { setHasNewChanges?: (has: boolean
...
-                </Box>
-                {!user && (
-                    <Typography sx={{ ml: '1rem', color: '#888', fontSize: '0.85rem' }}>
-                        Log in to save game options.
-                    </Typography>
-                )}
+                <Typography sx={{ mb: '1rem', color: '#aaa', fontSize: '0.9rem' }}>
+                    Customize the visibility of the game timer. These settings only affect what you see. Timers are always enabled in public games and always disabled in private games.
+                </Typography>
+                <FormControl component="fieldset">
```
AMMayberry1, line 288: I think if we can make it fit, it would be good here to show a preview of how the timer will look for each option. The images you've already used for the PR description would probably be fine, I'm just thinking that it will show a preview either to the right of the radio buttons or below them which will update when you pick a different option.

Just thinking that players probably don't intuitively understand the terms "turn time" and "main time" to the degree that we do, so the extra clarity may be helpful for those who are experiencing anxiety around this so they know what they'll get

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/GameOptionsTab.tsx
```diff
@@ -92,28 +124,78 @@ function GameOptionsTab({ setHasNewChanges }: { setHasNewChanges?: (has: boolean
...
-                        disabled={!user}
-                    />
-                </Box>
-                {!user && (
-                    <Typography sx={{ ml: '1rem', color: '#888', fontSize: '0.85rem' }}>
-                        Log in to save game options.
-                    </Typography>
-                )}
+                <Typography sx={{ mb: '1rem', color: '#aaa', fontSize: '0.9rem' }}>
+                    Customize the visibility of the game timer. These settings only affect what you see. Timers are always enabled in public games and always disabled in private games.
```
AMMayberry1: Probably worth bolding or otherwise emphasizing the phrase "always enabled in public games" just to make it super clear

src/app/_components/Gameboard/_subcomponents/OpponentTray/GameTimer.tsx
```diff
@@ -85,9 +90,14 @@ const GameTimer: React.FC = ({ ...props }) => {
     // Show opponent's main time in the circle when they are active on main time and player is not active
     const showOpponentMainTime = !playerIsActive && opponentIsActive && !isTurnTime;
 
+    if (timerVisibility === TimerVisibility.HideAll) {
```
AMMayberry1, line 95: Open question - I'm wondering if we should show some placeholder image or text here to remind the player that the timer still exists and is running, it's just hidden?

There are pros and cons in my mind. The pro is that players won't disable this and then totally forget it's there or mistakenly think they've disabled it completely, although that's also potentially resolved by my comment on the BE repo about adding softer warning indicators for low main time.

The con is that it might trigger that same anxiety response for players as actually seeing the running timer and continue to frustrate people, I'm not totally sure.

Good to go either way on it I think, especially if we add the soft warnings. Just wanted to float the idea.

AMMayberry1:
I'll raise this in chat as well but currently this option isn't available to anon users. I would kind of prefer that we make it available to them for accessibility reasons, but let's discuss.

AMMayberry1 (approved):

PR #681, opened 2026-05-18
Add storage support for all card languages

AMMayberry1:
This PR is a refactor of the image upload and image retrieval code to support having per-locale folder paths in our card image storage. Users cannot yet select a different language than EN, but we are migrating to using the new storage format for EN cards and testing that it works correctly before adding the option to change language.

### How to Test
The card images **have already been uploaded to our real card storage using the script**. The new folder structure does not overlap with the old one (see details below), so no data has been overwritten or removed.

This feature can be tested with EN cards by just spinning up the FE in local dev like normal. Other languages can be tested locally be setting the env var `NEXT_PUBLIC_CARD_IMAGE_LOCALE` to another value such as `fr` or `de`.

### Image sourcing
The locale images are pulled from the FFG API. These images are unfortunately lower-quality than what is hosted in SWUDB, but there is no equivalent source for images in all locales sadly.

Also, during preview season it is usually the case that images are available in swudb and not in the FFG API. In these cases, the script will use the EN image as a substitute so that the user has something to see for the card. On a future run, once the locale images have become available they will automatically be uploaded to replace the EN images.

### S3 layout: locale-aware folder structure
Card and token images are now stored under a per-locale subfolder so we can serve localized art (FR/DE/ES/IT) alongside English from the same bucket.

Before

```
cards/{SET}/{format}/{size}/{NNN}.webpcards/_tokens/{format}/{id}.webp
```

After
```
cards/{SET}/{locale}/{format}/{size}/{NNN}.webpcards/_tokens/{locale}/{format}/{id}.webp
```

Where:

`{locale} ∈ en, fr, de, es, it`
`{format} ∈ standard, truncated`
`{size} ∈ large, small (cards only — tokens have no size split)`


src/app/_utils/s3Utils.ts
```diff
@@ -39,7 +77,7 @@ export function s3CardImageURL(
 
     const tokenIds = ['3941784506', '3463348370', '7268926664', '9415311381', '8752877738', '2007868442', '6665455613']
     if (cardType?.includes('token') || (card.id && tokenIds.includes(card.id))) {
-        return s3ImageURL(`cards/_tokens/${format}/${card.id}.webp`);
+        return s3ImageURL(`cards/_tokens/${locale}/${format}/${card.id}.webp?v=3`);
```
danbastin, line 80: Technically unnecessary here since the path change will be sufficient to trigger a redownload of the asset. No harm in doing the bump though cause even `en` users will end up redownloading.

danbastin (approved):
Small comment on the image fetch URL that don't relaly need to be addressed. Fine as is. 

Vaya con dios on the python download script, but ultimately pretty low risk. As long as images land in the right directory it seems fine.

PR #688, opened 2026-05-20
Add cooldown to some actions to avoid double click pitfalls

### LINKED ISSUE #349: Add 0.5s delay on Claim Initiative and Pass buttons to help prevent misclicks
Some players report accidentally clicking the Claim Initiative button because they were repeat-clicking the "Done" button from a previous prompt.

We had the same issue with the "Done" button when resourcing and added a short delay there. Let's do the same with Claim Initiative and Pass in the action window

santiagovazquez:
Fixes #349 

This PR adds a `cooldown` prop to the `PromptButton` component.

The main reason of this small "refactor" is trying to avoid calls to `setTimeout` by other functions at render time. Now the component knows if it should have cooldown before triggering any action and the the setTimeout function is being called by its own hook which has all the logic related to the cleanup on unmount. 

Video:

https://github.com/user-attachments/assets/ab346ceb-586c-452e-b5e9-c00108dfb82b

Any feedback is welcome

### DIFF
--- a/src/app/_components/Gameboard/_subcomponents/PlayerTray/CardActionTray.tsx
+++ b/src/app/_components/Gameboard/_subcomponents/PlayerTray/CardActionTray.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import { Button, Box } from '@mui/material';
 import Grid from '@mui/material/Grid2';
 import { useGame } from '@/app/_contexts/Game.context';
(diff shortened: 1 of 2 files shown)


src/app/_components/Gameboard/_subcomponents/PlayerTray/CardActionTray.tsx
```diff
@@ -148,12 +148,15 @@ interface IButtonsProps {
     disabled?: boolean;
 }
 
+// disabling certain actions briefly to avoid double clicks:
+// - for a resource prompt, we disable the "done" button
+// - for the action window, we disable the "claim initiative" button
+const hasCooldown = (buttonType: string, prompType: string) => (buttonType === 'done' && prompType === 'resource') || (buttonType === 'claimInitiative' && prompType === 'actionWindow')
```
AMMayberry1, line 154: The issue mentions adding the delay for the Pass button as well and I think that makes sense. If the previous prompt also had two buttons it would definitely be possible to double-click into Pass on accident

PR #698, opened 2026-05-23
Fix popup from breaking on mobile

santiagovazquez:
## Change

Same behavior as in `src/app/_components/_sharedcomponents/Preferences/PreferencesComponent.tsx:32`

Width now mimics the logic from the sidebar.

Also, now we set the min width to the modal so it could be larger if the content needs it. 

## Current behavior

Mobile portrait broken

Mobile landscape not that bad

localhost_3000_GameBoard(iPhone 12 Pro)

Currently, no issue on desktop
3localhost_3000_GameBoard

## New behavior
pr_localhost_3000_GameBoard(iPhone 12 Pro)
pr2_localhost_3000_GameBoard(iPhone 12 Pro)
pr3_localhost_3000_GameBoard


moyerr (approved):

danbastin:
Can we check that this doesn't cause any weird behavior for card selection popups like invisible hand or Endless Legions?

santiagovazquez:
@danbastin 
I tested both desktop and mobile portrait/desktop and I think it looks way better (it doesn't break like it breaks now).

### New behavior

localhost_3000_GameBoard(iPhone 12 Pro) (2)

localhost_3000_GameBoard(iPhone 12 Pro)


### Old behavior

old_localhost_3000_GameBoard(iPhone 12 Pro) (4)

localhost_3000_GameBoard(iPhone 12 Pro) (4)

danbastin (approved):

PR #699, opened 2026-05-23
Avoid username to hide the leader and give more visibility for mobile…

### LINKED ISSUE #690: Long Usernames block leader portrait on mobile
Image

santiagovazquez:
… devices

fixes #690 

## Change
I tried to maximize the leader visibility so I removed the background (if you use the background also you'd need paddings to make it look better) and added some text shadow. Also now it has text overflow ellipsis so the name will be cut if it's too long.

## Current behavior
mobile_portrait_no_changes
mobile_landscape_no_changes
desktop_no_changes

## New behavior
Desktop remains unchanged

desktop_with_changes

Mobile

localhost_3000_(iPhone 12 Pro) (2)


src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
```diff
@@ -326,12 +326,14 @@ const LeaderBaseCard: React.FC<ILeaderBaseCardProps> = ({
...
-            bottom: '0',
+            bottom: { xs: '-2px', md: '0' },
             left: '50%',
             transform: 'translateX(-50%)',
-            backgroundColor: 'black',
             borderRadius: '0.5rem 0.5rem 0 0',
-            p: '5px 10px',
+            p: { xs: 0, md: '5px 10px' },
+            maxWidth: { xs: '100%', md: 'none' },
+            backgroundColor: { xs: 'transparent', md: 'black' },
```
danbastin: I think restricting to 1 line and the font size fix is sufficient here. Lets maintain the back plate for now and we can revisit when we take a closer look at the center column.

src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
```diff
@@ -345,6 +347,9 @@ const LeaderBaseCard: React.FC<ILeaderBaseCardProps> = ({
             color: 'white',
             fontWeight: '600',
             fontSize: '1em',
+            textOverflow: 'ellipsis',
+            textWrap: { xs: 'nowrap', md: 'wrap' },
```
danbastin: Lets jsut make this `nowrap` for everything. It shouldn't matter cause I tink we don't allow spaces in usernames so this really only affects anon users for now.

src/app/_components/_sharedcomponents/Cards/LeaderBaseCard.tsx
santiagovazquez: let me change it. I removed it because I tested in small screen devices and the cards sometimes are insanely small so any gain in card visibility was a plus but I agree it doesn't look better.

danbastin (approved):

PR #709, opened 2026-05-25
Create dedicated timer tutorial popup

AMMayberry1:
Migrated the timer explanation popup from being an announcement to a dedicated tutorial popup that will happen the first time a user tries to queue or join a lobby. This is already how we handle the undo tutorial, so we're just extending the pattern.

BE PR: https://github.com/SWU-Karabast/forceteki/pull/2431

image


santiagovazquez:
I really like this. Great stuff

danbastin (approved):

Issue #663, opened 2026-04-03
Create a stepper component for number selection prompts


moyerr:
Using the dropdown for number selection can be a pretty frustrating UX, especially for frequent interactions like with Credit tokens. Interacting with the text box automatically pops up the keyboard on Mobile/Tablet, when it shouldn't be necessary.

We could improve this experience with some kind of stepper component that could be incremented/decremented with buttons, and add support for keyboard arrows to adjust the values too.

Many cards could benefit from this enhancement such as:

- Force Lightning
- SOR C-3PO
- Emergency Powers
- etc...

PR #712, opened 2026-05-26
Migrate to MUI 7

santiagovazquez:
I went through carefully [the migration guide](https://mui.com/material-ui/migration/upgrade-to-v7/) to update the MUI version.  

# Why migrate?

I would like to work on #663 and use the new `NumberField` MUI component from `v7`
Screenshot 2026-05-26 at 8 57 30 AM

Also:
 - Old versions tie you to old version of other libraries
 - Be able to use newer stuff and the new docs
 - Vulnerabilities

### DIFF
--- a/src/app/DeckPage/layout.tsx
+++ b/src/app/DeckPage/layout.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import { Box, Typography } from '@mui/material';
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
 import { s3ImageURL } from '@/app/_utils/s3Utils';
 
 export default function DeckLayout({
--- a/src/app/Unimplemented/layout.tsx
+++ b/src/app/Unimplemented/layout.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import { Box, Typography } from '@mui/material';
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
 import { s3ImageURL } from '@/app/_utils/s3Utils';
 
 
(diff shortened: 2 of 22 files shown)


src/app/DeckPage/layout.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import { Box, Typography } from '@mui/material';
```
danbastin, line 2: Is this split necessary? We're still decomposing the import above so we shouldn't need to change here.

src/app/mod/page.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
 import { getServerSession } from 'next-auth';
 import { authOptions } from '@/app/_utils/auth';
-import { Box, Alert } from '@mui/material';
```
danbastin, line 4: Same comment

src/app/Unimplemented/layout.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import { Box, Typography } from '@mui/material';
```
danbastin, line 2: 1 more

src/app/DeckPage/layout.tsx
santiagovazquez: Yes, it's an issue with the lastest version of the `7` MUI version as [stated here](https://github.com/mui/material-ui/issues/46688#issuecomment-3660091201). It was failing at build time for me and the change fixed it. Once we have this version out there we could bump to the next one.

santiagovazquez: I'm not saying we should change it, we could try to revert the change in the next version. As an FYI, they recommend to avoid barrel imports for faster development times
Screenshot 2026-05-26 at 4 28 16 PM

danbastin (approved):

PR #719, opened 2026-05-29
Add css loader instead of a video

santiagovazquez:
On iphone when joining a match, it loads a full screen video. This PR is to remove the `mp4` loader and use a css pure version of it.

https://github.com/user-attachments/assets/a251517c-7d81-46e7-95cc-77e1e87324a7


src/app/_components/QuickGame/SearchingForGame/MatchLoader.tsx
```diff
@@ -0,0 +1,76 @@
+import { Box, SxProps, Theme } from '@mui/material';
+import { SystemStyleObject } from '@mui/system';
+
+const styles = {
+    container: {
```
danbastin, line 5: Might need a little padding here to accomodate the image scaling to `1.15`. Right now you can see the top of the image get cut off a bit.

santiagovazquez: sure, let me add it and I'll record another video

danbastin (approved):

PR #720, opened 2026-05-29
UI for choosing card language

AMMayberry1:
Officially adding the FE support for card images in all languages! See example images below for relevant UI changes. Here are the major changes:

#### Major Changes
- Added a dropdown selection for language in the "Game Options" tab in the Preferences page. The option exists for both logged-in and anonymous users (similar to sound settings)
- Added a small globe icon in the top-right next to the socials as a visual hint that this option is available. This will link directly to the Game Options tab
- The new account popup now has an inline option to change the card language to help increase awareness
- Updated the announcement popup to let everyone know

#### Issues with Preview Images
Since swudb has its images (always English) available before the FFG API, there will be a delay before the newest previews are available in card locales. 

To address this, the upload script will set a shorter cache period (one week) for images in the preview set so that browsers will refresh on a regular cadence and pick up the locale images after they become available. The preview set setting is hardcoded into the script and will need to be updated periodically.

To help make users aware of this issue, we're adding a popup that appears when selecting a card language other than English to help explain the situation.

# Previews

## Preference dropdown
image

## Notice popup
image

## Language selection button on control panel
image

## Change to new account popup
image


src/app/_contexts/CardImageLocale.context.tsx
```diff
@@ -0,0 +1,82 @@
...
+                gameOptions: {
+                    ...current.gameOptions,
+                    cardLanguage: next,
+                },
+            });
+        }
+    }, [user]);
+
+    return (
+        <CardImageLocaleContext.Provider value={{ locale, setLocale }}>
```
danbastin: Memoize this object and only have it update on `locale` and `setLocale`. in general we never want to pass an un-memoized object into a context provider since it isn't referentially stable.

src/app/_utils/s3Utils.ts
```diff
@@ -35,41 +36,29 @@ const isTokenCardId = (id: string | undefined): boolean =>
...
+    [CardImageLocale.English]: 'English',
+    [CardImageLocale.French]: 'Français',
+    [CardImageLocale.German]: 'Deutsch',
+    [CardImageLocale.Spanish]: 'Español',
+    [CardImageLocale.Italian]: 'Italiano',
+};
 
 export function s3CardImageURL(
     card: ICardData | ISetCode | IServerCardData | IPreviewCard,
+    locale: CardImageLocale,
```
danbastin, line 59: Should we set an english default here?

AMMayberry1: I think not, managing the default is handled elsewhere so it could just hide potential bugs to give a default here

danbastin (approved):

PR #721, opened 2026-05-29
Add support for numerical prompts

santiagovazquez:
Before we can merge this, we must merge the [BE part](https://github.com/SWU-Karabast/forceteki/pull/2464)

Fix for https://github.com/SWU-Karabast/forceteki-client/issues/663

Showcase:
Screenshot 2026-05-29 at 4 02 16 PM
numeric_input_mobile


src/app/_components/_sharedcomponents/Popup/PopupVariant/NumberPopup.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+        return (
+            <>
+                {data.description && (
+                    <RichText text={data.description} sx={textStyle} component={Typography}/>
+                )}
+
+                <NumberSpinner
+                    min={data.min}
+                    max={data.max}
```
AMMayberry1, line 81: It looks like if you manually type in a value that's too large and press "Done", it will jump to the highest legal value. Is there an option to instead block and give an error or something to avoid accidental inputs?

AMMayberry1 (approved):

PR #733, opened 2026-06-03
Disable Join Queue action if conditions are not met

santiagovazquez:
## Goal

Don't let the user hit "Join Queue" when conditions are not met to avoid showing popup errors.

What's in this PR:
 - The user shouldn't be able to join if decks are still loading
 - The user shouldn't be able to join if he selects new deck but input is empty
 - Input for new decks now triggers a validation at the `onBlur` event.

## Problem

Hit Join before decks are loaded

https://github.com/user-attachments/assets/f7c7a752-2fdb-46b8-a986-9fd3bf66aa85

Hit join without any input

https://github.com/user-attachments/assets/c6407c44-65c4-40fd-8382-e7f3db407790

## Solution

Loading fix (button disabled until the decks are loaded)

https://github.com/user-attachments/assets/498a3776-a7c1-448b-b978-bc259b73d7b1

Empty input fix

https://github.com/user-attachments/assets/587d35bf-3b1e-4dac-bf9b-b1ff872aa547


danbastin:
We whould probably mirrir the black deck field for the drop down if they have nothing selected there. Otherwise I'm all for this.

santiagovazquez:
> We whould probably mirrir the black deck field for the drop down if they have nothing selected there. Otherwise I'm all for this.

Not sure if I follow. What should I mimic?

danbastin (approved):

PR #746, opened 2026-06-14
stop propagating clicks from subcard element if is clickable

Ravarix:
Shield widgets render on top of their parent card, which is itself clickable. 
Without stopping propagation, a single click on a selectable shield also bubbles to the card's onClick and emits a second cardClicked for the underlying unit. 
With Alliance Outpost that buffered second click auto-applies the buff to the token whose shield was just defeated.

### DIFF
--- a/src/app/_components/_sharedcomponents/Cards/GameCard.tsx
+++ b/src/app/_components/_sharedcomponents/Cards/GameCard.tsx
@@ -226,8 +226,14 @@ const GameCard: React.FC<IGameCardProps> = ({
         (onClick || defaultClickFunction)();
     }
 
-    const subcardClick = (subCard: ICardData) => {
+    const subcardClick = (event: React.MouseEvent, subCard: ICardData) => {
         if (subCard.selectable) {
+            // Shield widgets render on top of their parent card, which is itself
+            // clickable. Without stopping propagation, a single click on a selectable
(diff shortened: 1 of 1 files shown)


src/app/_components/_sharedcomponents/Cards/GameCard.tsx
```diff
@@ -226,8 +226,14 @@ const GameCard: React.FC<IGameCardProps> = ({
         (onClick || defaultClickFunction)();
     }
 
-    const subcardClick = (subCard: ICardData) => {
+    const subcardClick = (event: React.MouseEvent, subCard: ICardData) => {
         if (subCard.selectable) {
+            // Shield widgets render on top of their parent card, which is itself
```
danbastin: Can we make this comment a little more generic? This propagation stop seems applicable to any token we might drop on the card that ends up being part of a similar double action.

Ravarix: I think only shield actually renders over the card. XP/advantage are attached below.

danbastin: For now yes. But this same logic will apply to any other token/clickable that might get rendered on the card border in the future. Just want to make it clear this isn’t just a shield specific fix even though it’s currently the only relevant target right now.

danbastin (approved):

PR #760, opened 2026-06-26
Move remaining deckbuilder calls from Amplify to client

AMMayberry1:
The SWUDB and Melee.gg deck fetchers have been migrated to the BE to handle CORS issues. The remaining deckbuilders needed to be moved to the client so that we are no longer paying for Amplify to handle our deckbuilder calls.

This PR migrates the remaining deckbuilder fetches from Amplify to the client code, and also majorly refactors the entire pattern. Each deckbuilder integration is now encapsulated into a single file with a handful of required fields.

This includes the "backend-fetching" deckbuilders swudb and melee, so that everything can use the same basic pattern.

I have tested with all of the deckbuilder sites and they all work except for KyberDecks, which seem to be having issues on their end and are currently not working even in prod. I added a `hiddenFromPublicList` option to providers which will keep the functionality but hide them from the list of supported deckbuilders, KyberDecks is hidden until the devs can fix the issues.


moyerr (dismissed):

src/app/_utils/deckProviders/core/httpGetJson.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export async function httpGetJson(url: string, providerName?: string): Promise<Response> {
+    const isDev = process.env.NODE_ENV === 'development';
+    const requestUrl = isDev ? `/api/deck-proxy?url=${encodeURIComponent(url)}` : url;
+
+    let response: Response;
+    try {
+        response = await fetch(requestUrl, {
+            method: 'GET',
+            cache: 'no-store',
```
danbastin, line 32:
```suggestion
            cache: 'no-store',
            signal: AbortSignal.timeout(5000),
```

There is a default time for fetch requests but they differ per browser and are generally pretty long (at least 90 seconds). Just in case anything gets wonky we can time this out after 5 seconds which should be plenty for these requests

moyerr (approved):

PR #761, opened 2026-06-26
Only get active cosmetics from the BE for rendering on game board

AMMayberry1:
BE PR: https://github.com/SWU-Karabast/forceteki/pull/2605

Currently, the BE sends the entire list of all available cosmetic options (cardbacks, backgrounds) to the FE when a user first hits the site. From this, the FE chooses the active customization options for the user based on their account settings.

This is incredibly inefficient from a data perspective, since we should be sending just the customizations that are active for the user instead of the entire catalog. This PR streamlines the data contract so that we can do exactly that.

To test these changes, use the wiki instructions here: https://github.com/SWU-Karabast/forceteki/wiki/Local-S3-Setup

### DIFF
--- a/src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/CosmeticsTab.tsx
+++ b/src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/CosmeticsTab.tsx
@@ -30,9 +30,14 @@ function CosmeticsTab() {
                 /* setSelectedPlaymat(cosmetics.playmat ?? 'none');
                 setDisablePlaymats(cosmetics.disablePlaymats ?? false);*/
             }
-            fetchCosmetics();
         }
     }, [user]);
+
+    useEffect(() => {
+        if (user?.id) {
+            fetchCosmetics();
+        }
+    }, [user?.id, fetchCosmetics]);
 
     const onCardbackClick = async (is: string) => {
         try {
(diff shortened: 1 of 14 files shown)


moyerr (dismissed):

src/app/mod/subpages/CosmeticsManagerTab.tsx
CheBato, line 304: Just a suggestion, while testing i noticed we keep fetchingCosmetics everytime we change our cosmetic, do you think we could remove this to lower traffic everytime someone changes their cosmetic? Edit: I'm not really sure what exactly it would influence. Maybe that users won't have their cardbacks super updated everytime?

AMMayberry1: It's being run after any change so that the mod page will immediately reflect the update and be able to see the updated list

CheBato: Ah my mistake. I commented on the wrong line since the fetch cosmetics is also in the **CosmeticsTab** thats where i thought it wouldn't be necessary to update the cosmetics everytime someone selects a cardback.

src/app/_components/_sharedcomponents/Preferences/PreferencesSubElementVariants/CosmeticsTab.tsx
CheBato: This is the one i had in mind. Basically what happens is that everytime we change the users cardback or background it also fetches the cosmetics. I was thinking if we could potentially move this so that it triggers only once. Let me know if this makes sense.

moyerr (approved):

PR #763, opened 2026-06-30
Track username change history in mod page

AMMayberry1:
BE PR: https://github.com/SWU-Karabast/forceteki/pull/2609

As a first step towards improving the username management tools in the mod page, we are adding support for tracking username changes as first-class citizens in the DB.

In the UI, the following will now be tracked in the history:
- Initial username state - either for new accounts or accounts that existed before this change which will be migrated to the new system
- Manual username changes (not due to mod action)
- Force renames will now show the before / after for the name change

See BE PR for instructions on how to test.

Example with migrated user:

image


src/app/mod/subpages/UserManagementTab.tsx
```diff
@@ -304,6 +310,7 @@ const UserManagementTab: React.FC = () => {
 
     };
     // ==================== Render ====================
+    const history = selectedPlayer ? buildUserHistory(modActions, usernameChanges) : [];
```
CheBato: buildUserHistory seems to fire up on every render and also when i type in the notes. Could we move it to a memo? That way it could fire up only when modActions and usernameChanges?

src/app/mod/subpages/UserManagementTab.tsx
CheBato, line 145: we may need to add 
```ts
setUsernameChanges(refreshed.usernameChanges || []);
```
here since otherwise the refresh won't get whether a person did a usernameChange when someone does an action.
This probably won't happen so often but might be cool having it refresh here.

CheBato, line 175: same thing as comment above https://github.com/SWU-Karabast/forceteki-client/pull/763/changes#r3514823572

moyerr (approved):

PR #805, opened 2026-08-08
Require a deck selection before enabling Join Queue / Create Game

### LINKED ISSUE #743: Create Game button active with empty saved deck dropdown
Image

Current behavior: 

When No deck is selected with the saved deck dropdown the Join queue/Create game button is enabled. 

Expected Behavior: 

Join button should be disabled when no deck has been provided.

rshutton1:
The disabled check on both buttons only covered the New Deck path — it
looked at whether the deck link box was empty. In Saved Deck mode nothing
checked if you'd actually picked a deck, so the button stayed clickable
with an empty dropdown.

Added a check for that in both forms, plus a short line under the dropdown
saying why the button is greyed out. I went with plain grey text instead of
the red error styling, since you haven't really done anything wrong yet.

If you have no saved decks at all, the message points you at New Deck
instead. The dropdown's "No saved decks found" only shows up once you open
it, so with it closed you'd just get a blank field and a dead button with
no explanation.

Tested locally: with decks saved, with none saved, on both tabs, and
confirmed the New Deck path still behaves the same. I couldn't test the SWU
Stats deck source since I don't have an account linked.

Fixes https://github.com/SWU-Karabast/forceteki-client/issues/743


rshutton1:
Updating with screenshots to show various states between quick match and create lobby when missing a deck to play with.
deckselect1
ds2
ds3


cl1
cl2
cl3

danbastin (approved):

PR #808, opened 2026-08-16
Replace alert() calls with an in-app error screen

### LINKED ISSUE #756: Replace calls to "alert" with nicer error display
We should have a styled error pages/popups to display errors instead of:

Image

Also have a way to recover from the error: "Refresh", "Reconnect"

rshutton1:
image-1786915036938

Browser alert() dialogs were used for every error the client surfaced. They sit outside the app, can't be styled, and offer no way to recover beyond an OK button - alert(error) in the socket handler even dumped a raw error object at the user.

Add an ErrorScreenProvider that renders errors as a Star Wars Unlimited card on the site background, which flips over to reveal what went wrong. The card back reuses /card-back.png with the same treatment as the deck pile, so it reads as part of the game rather than a browser dialog.

Callers use showErrorScreen({ title, message, detail, actions }). Actions receive a dismiss callback, so each error can offer whatever recovery makes sense - returning to the main menu for fatal connection errors, retrying for failed fetches, dismissing for everything else. Raw server text goes in the detail field, under a readable sentence, instead of being the whole message.

The provider lives in ClientProviders so it wraps both the game board and the standalone pages. Game.context now uses it rather than holding its own error state.

Also clears the heartbeat interval in SearchingForGame before showing the error. The user now leaves via the error screen instead of being redirected immediately, so without this it would fire every second.

The card honours prefers-reduced-motion by skipping the flip, and moves focus to the first recovery action once the error is readable.

Fixes #756

### DIFF
--- a/src/app/DeckPage/page.tsx
+++ b/src/app/DeckPage/page.tsx
@@ -23,6 +23,7 @@ import {
     updateDeckFavoriteInLocalStorage,
 } from '@/app/_utils/ServerAndLocalStorageUtils';
 import { useUser } from '@/app/_contexts/User.context';
+import { useErrorScreen } from '@/app/_contexts/ErrorScreen.context';
 import { useSession } from 'next-auth/react';
 
 interface IDeckSummaryCardTileProps {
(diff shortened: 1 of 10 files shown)


rshutton1:
Additional screenshots as requested, 1 of the more serious variety (connection lost) and 2 minor (Couldnt load decks, couldnt join lobby/game)

image-1786988725406
image-1786988732120
image-1786988740061

danbastin:
Overall I like the hook and definitely prefer the styled alert over the default. The card flip is fun, but ultimately I don't think adds anything to the usefulness of the alert. I think we can just have the alert box pop up witht eh alert contents.

PR #814, opened 2026-09-05
Break down GameCard into smaller presentational components

santiagovazquez:
This PR breaks down GameCard into smaller presentational components as a first step toward separating rendering from game logic.

  - Extract stat badges, status icons, token badges, and the token stack layout.
  - Move card status checks into shared utilities.
  - Add Storybook galleries for the extracted components.
  - Align Storybook’s fonts, theme, and SVG handling with the app.
  - Extract card preview popover logic from the game card render.

  The goal is to make GameCard’s visual changes and states easier to test in isolation. This is an intermediate step: game interactions remain in GameCard, while the extracted components handle presentation.

### New stories

Screenshot 2026-09-05 at 7 39 38 PM

Screenshot 2026-09-05 at 7 39 44 PM

Screenshot 2026-09-05 at 7 39 32 PM


src/app/_components/_sharedcomponents/_styledcomponents/TokenContainer.tsx
```diff
@@ -269,8 +269,7 @@ export function TokenContainer({ type, stroke, sx, children, ...boxProps }: Toke
                 fontVariantNumeric: 'tabular-nums',
                 userSelect: 'none',
                 color: appearance.color,
-                ...sx,
-            }}
+            }, ...(Array.isArray(sx) ? sx : [sx])]}
```
danbastin, line 272: We'll probably be doing a good bit of this. Can we pull this into a utility that makes this a little more readable?

src/app/_components/_sharedcomponents/Cards/GameCard/cardHooks.tsx
```diff
@@ -0,0 +1,217 @@
...
+import { useLeaderCardFlipPreview } from '@/app/_hooks/useLeaderPreviewFlip';
+import { useLongPress } from '@/app/_hooks/useLongPress';
+import ThreeSixty from '@mui/icons-material/ThreeSixty';
+
+/**
+ * Based on the card positioning, return anchorOrigin and transformOrigin for the popover component
+ * @see https://mui.com/material-ui/api/popover/
+ * @param card
+ */
+export const usePopoverConfig = (card: ICardData): { anchorOrigin: PopoverOrigin, transformOrigin: PopoverOrigin } => {
```
danbastin, line 15: Should these live one level up since all cards will make use of this?

src/app/_components/_sharedcomponents/Cards/GameCard/cardHooks.tsx
```diff
@@ -0,0 +1,217 @@
...
+            onMouseEnter: handlePreviewOpen,
+            onMouseLeave: handlePreviewClose,
+            'data-card-url': cardUrl,
+            'data-card-type': cardType,
+            'data-card-id': cardId,
+            ...longPressHandlers
+        }
+    };
+    const popover = disabled ? null : (
+        <Popover
```
danbastin, line 171: Thinking about the future of a hook like this. All of our cards across the site use this popover but only 1 popover is needed at a time. What do we think of moving the rendered portion to a context provider and this hook then only handles the popup state and props. We reduce the rendered popup to 1 (we only ever display 1) and consumers would tell the popover how it should render when that card is hovered. 

I think this would clean up a ton of popover code we have scattered around the site. Not a blocker here since thats a bit out of scope for this PR.

src/app/layout.tsx
```diff
@@ -2,21 +2,9 @@
 import type { Metadata, Viewport } from 'next';
 import React from 'react';
 import ClientLayout from './ClientLayout';
-import { Barlow, Barlow_Semi_Condensed } from 'next/font/google';
+import { barlow, barlowSemiCondensed } from './_theme/fonts';
```
danbastin, line 5: 👍

danbastin (approved):

src/app/_components/_sharedcomponents/Cards/GameCard/cardHooks.tsx
santiagovazquez: I'm totally on board with this. Once we merge this, I can take that one as the first follow-up

src/app/_components/_sharedcomponents/_styledcomponents/TokenContainer.tsx
santiagovazquez: I know it looks dirty but it's the only way to extend the current behavior of the `sx` prop (allowing both `array` or `object` as parameter). I'll make a utility for this and replace all the occurrences as a follow-up

src/app/_components/_sharedcomponents/Cards/GameCard/cardHooks.tsx
santiagovazquez: I put it in there since the only consumer was the `GameCard` component. I believe this will be part a function only called by a new global `CardDetailsPopover` rendered by a new content in some follow-up PR.

PR #819, opened 2026-09-18
(fix) mobile: trigger active effects on longpress

santiagovazquez:
## Bug

When trying to read the active effect on mobile the popover keeps closing.

In this PR we're trying to do the same as Cards:
- on mobile: show preview on long press until next interaction
- on desktop: show preview on hover

## Fix

https://github.com/user-attachments/assets/e3750d7f-5198-48ee-be6f-c835d0afe273


santiagovazquez:
@CheBato I need your help for reviewing & proper QAing. I did my QA and it seems to be working fine. Let me know if you think of different scenarios we could test.

danbastin (approved):

